PR #9, opened 2017-07-12
fix: avoid collisions with parameters and user namespace

adparker:
* added a super simple test
* switched from kwargs to parameters to avoid a conflict between notebook parameters and user namespace

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="VcsDirectoryMappings">
-    <mapping directory="$PROJECT_DIR$" vcs="Git" />
-  </component>
-</project>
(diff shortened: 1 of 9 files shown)


rgbkrk:
Sweet! How should this get invoked for Travis?

rgbkrk:
You can add the testing line in the `script` section for Travis to run the test: https://github.com/nteract/papermill/blob/997bbda3c24ee4b9863344645cc5e093cd77ddb6/.travis.yml#L8

rgbkrk:
You'll also want to `pip install pytest` in the install, so the travis yaml will look like:

```yaml
language: python
python:
  - "2.7"
  - "3.5"
# command to install dependencies
install:
  - pip install pytest
  - pip install -e .
# command to run tests
script: py.test
```

rgbkrk:
After this we'll certainly put out a patch release since this wasn't fully working yet. :wink:

rgbkrk:
Ok -- issue we run into here -- the notebook is python2 and nbconvert won't fallback to python3.

adparker:
Very hacky to get this to work in both Python 3 and Python 2. Sad.

adparker:
PTAL: @rgbkrk and @ewmassey

rgbkrk:
Cool, I'll go ahead and bring this in.

rgbkrk:
Thanks @adparker!

PR #10, opened 2017-07-13
fetch data and dataframes

ewmassey:
All notebook data dict will be returned from pm.fetch_notebook_data(nb) and a collection of notebooks can return a dataframe from pm.fetch_notebooks_dataframe(nbs).

### DIFF
--- a/papermill/api.py
+++ b/papermill/api.py
@@ -1,6 +1,10 @@
 import io
+import os
+
 import nbformat
+import pandas as pd
 from IPython.display import display
+from six import string_types
 
 
 RECORD_OUTPUT_TYPE = 'application/papermill.record+json'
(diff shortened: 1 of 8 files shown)


papermill/api.py
```diff
@@ -140,43 +161,51 @@ def record_value(name, value):
...
     """
-    if not path.startswith('s3://'):
-        raise ValueError('path must start with s3://')
-    path = path[5:]
-    if path:
-        if '/' in path:
-            return path.split('/', 1)
-        else:
-            return [path, '']
+    if isinstance(notebooks, basestring):
```
rgbkrk: Use the string handling for both Python 2 and Python 3 from the other PR so this supports both. I'll help fix the build in the other PR in the morning.

rgbkrk:
You'll need to rebase this branch to get this merged in. Additionally, you'll want to move the requirements to `setup.py` (see what's on master) rather than the requirements file.

PR #15, opened 2017-07-28
Link to bindings for recording values

rgbkrk:
P.S. I miss Markdown

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,6 +17,15 @@ Installation
   pip install papermill
 
 
+In-Notebook bindings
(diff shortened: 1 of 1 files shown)


rgbkrk:
I'm not sure how to refer to the bindings, since papermillr (for R) is only for recording papermill values and metadata (reading the notebook is still done in python).

ewmassey:
People can always add in the ability to read in the notebooks from R later.
Calling them bindings makes sense to me.

PR #16, opened 2017-08-02
Removing reader/writer customization

ewmassey:
Removing reader and writer customizing framework in favor of monkey patching for now.

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -8,7 +8,7 @@ import nbformat
 
 from papermill.conf import settings
 from papermill.exceptions import PapermillException
-from papermill.iorw import read_notebook, save_notebook
+from papermill.iorw import load_notebook_node, write_ipynb
 
 from six import string_types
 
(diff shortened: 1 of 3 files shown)


papermill/execute.py
```diff
@@ -44,8 +44,8 @@ def _parameterize_notebook(notebook_path, output_path, parameters):
             val = '"%s"' % val  # TODO: Handle correctly escaping input strings.
         param_content += '%s = %s\n' % (var, val)
 
-    # Remove the first cell of the block and replace it with the param content.
-    nb = read_notebook(notebook_path)
+    # Find `parameters` cell and replaced with our parameters.
```
rgbkrk: ...and replace with...

ewmassey:
I had two functions called read_notebook. One read a ipynb file and
returned a NotebookNode. The other is the API function that returns a
Papermill object. I removed the former to avoid confusion.

papermill/execute.py
ewmassey: I had two functions called read_notebook. One read a ipynb file and
returned a NotebookNode. The other is the API function that returns a
Papermill object. I removed the former to avoid confusion.

rgbkrk: My comment here was to fix the grammar. :)

PR #17, opened 2017-08-04
Load from yaml

ewmassey:
Execute_notebook can read from a file or from a dict.
Moved listing of notebooks into a directory to make it easier to monkey patch just that functionality.


rgbkrk:
Since there was a merge commit I went ahead and squash merged this.

PR #23, opened 2017-08-21
Changed papermill_io to a file handler framework.

ewmassey:
Using a File Handler framework to simplify custom file handling.


rgbkrk:
Awesome, this is looking good. I'm happy to see the simplified S3 API here, built on the boto3 session.

rgbkrk:
Should boto3 be optional or do we want to enforce that?

You know what, nevermind. We should go with opinionated and simple.

PR #26, opened 2017-08-22
Live Notebook updates

ewmassey:
Writes the notebook as each cell executes to mimic the same functionality as "run all" in the web ui. Also includes timing and success status for each cell. Give this PR a little extra attention since a fair amount about the execution is changing.

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -17,7 +17,8 @@ from six import string_types
 
 def preprocess(self, nb, resources):
     """
-    This function is for monkey patching the nbconvert Preprocessor.preprocess method.
+    This function monkey patches the Preprocessor.preprocess method.
+
     We are doing this for the following reasons:
     1. Notebooks will stop executing when they encounter a failure but not raise a CellException.
        This allows us to save the notebook with the traceback even though a CellExecutionError
@@ -25,16 +26,7 @@ def preprocess(self, nb, resources):
 
     2. We want to write the notebook as cells are executed. We inject our logic for that here.
 
-    3. We will be included cell execution metrics.
-
-    Original Docstring:
-
-    Preprocessing to apply on each notebook.
-
-    Must return modified nb, resources.
-
-    If you wish to apply your preprocessing to each cell, you might want
-    to override preprocess_cell method instead.
+    3. We want to include timing and execution status information with the metadata of each cell.
 
     Parameters
     ----------
@@ -62,11 +54,10 @@ def preprocess(self, nb, resources):
         cell.outputs = []
 
     # Execute each cell and update the output in real time.
-    writers = []
     with futures.ThreadPoolExecutor(max_workers=1) as executor:
         for index, cell in enumerate(nb.cells):
             cell.execution_count = "*"
-            writers.append(executor.submit(write_ipynb, nb, output_path))
+            future = executor.submit(write_ipynb, nb, output_path)
             t0 = datetime.datetime.utcnow()
             try:
                 nb.cells[index], resources = self.preprocess_cell(cell, resources, index)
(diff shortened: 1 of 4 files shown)


papermill/execute.py
```diff
@@ -13,26 +15,75 @@
...
-class PapermillExecutePreprocessor(ExecutePreprocessor):
+def preprocess(self, nb, resources):
+    """
+    This function is for monkey patching the nbconvert Preprocessor.preprocess method.
+    We are doing this for the following reasons:
+    1. Notebooks will stop executing when they encounter a failure but not raise a CellException.
+       This allows us to save the notebook with the traceback even though a CellExecutionError
+       was encountered.
+
+    2. We want to write the notebook as cells are executed. We inject our logic for that here.
```
rgbkrk: Superb, looking forward to reviewing this whole thing.

papermill/execute.py
```diff
@@ -13,26 +15,75 @@
...
-    def preprocess(self, nb, resources):
-        """
-        The default preprocess behavior if allow_errors = True is to continue executing all the cells.
-        We want the notebook to cease execution and then write out the state of the notebook with the traceback
-        on the cell that failed.
-        """
-        # Clear out all outputs prior to execution.
-        for cell in nb.cells:
-            cell.outputs = []
+    3. We will be included cell execution metrics.
```
rgbkrk: What is this meant to say?

papermill/execute.py
```diff
@@ -13,26 +15,75 @@
...
 
-        # Catch CellExecutionError if thrown which lets this method finish so we can write the state of the notebook
-        # to disk.
-        try:
-            nb, resources = super(PapermillExecutePreprocessor, self).preprocess(nb, resources)
-        except CellExecutionError:
-            pass
+    Original Docstring:
```
rgbkrk: I'd either cut out this line or make this all fit together in one narrative for the docstring (can be broken up in paragraphs as needed).

papermill/execute.py
```diff
@@ -13,26 +15,75 @@
...
+            duration=None
+        )
+        cell.outputs = []
+
+    # Execute each cell and update the output in real time.
+    writers = []
+    with futures.ThreadPoolExecutor(max_workers=1) as executor:
+        for index, cell in enumerate(nb.cells):
+            cell.execution_count = "*"
+            writers.append(executor.submit(write_ipynb, nb, output_path))
```
rgbkrk: These futures could come in any order -- we should not let writing happen out of order here. Either each write should be queued up (possibly clearing the queue if executing is faster than the writing to disk) or wait for the result of the future `future.result()`.

rgbkrk: Adding on to this, is anything done with the list of futures (`writers`)?

ewmassey: Nope. Going with your original suggestion.

papermill/execute.py
ewmassey: cleaning up docs on next push

papermill/execute.py
rgbkrk: If you go with the `future.result` part, you may realize you can simplify even further and not make a thread pool (just write to disk at the end of the loop each time).

ewmassey: I want to write asynchronously to not add more time to running a cell than I have to.

rgbkrk: Wouldn't you do it after the cell timing though?

ewmassey:
That would still make the time between cell executions be kernel time +
write time. I want max(kernel, write).

papermill/execute.py
```diff
@@ -13,26 +15,67 @@
...
+                cell.metadata['papermill']['exception'] = False
+            except CellExecutionError:
+                cell.metadata['papermill']['exception'] = True
+                break
+            finally:
+                future.result()
+                t1 = datetime.datetime.utcnow()
+                cell.metadata['papermill']['start_time'] = t0.isoformat()
+                cell.metadata['papermill']['end_time'] = t1.isoformat()
+                cell.metadata['papermill']['duration'] = (t1 - t0).total_seconds()
```
rgbkrk: You should move `future.result()` after your time calculations, otherwise you're also measuring IO. This will become pretty glaring with larger notebooks.

PR #27, opened 2017-08-26
Fixed failing tests

### LINKED ISSUE #25: Notebook validation failed when output with markdown
If I add a markdown cell, for example 

`# This is the title`

After execute from papermill, it will report

qq 2

I think the reason might be that papermill has not dealt with `markdown cell` yet？

charsmith:
This fixes #25 and both failing tests.

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -51,12 +51,14 @@ def preprocess(self, nb, resources):
             end_time=None,
             duration=None
         )
-        cell.outputs = []
+        if hasattr(cell, "outputs"):
+            cell.outputs = []
 
     # Execute each cell and update the output in real time.
     with futures.ThreadPoolExecutor(max_workers=1) as executor:
         for index, cell in enumerate(nb.cells):
-            cell.execution_count = "*"
+            if hasattr(cell, "execution_count"):
+                cell.execution_count = "*"
             future = executor.submit(write_ipynb, nb, output_path)
             t0 = datetime.datetime.utcnow()
             try:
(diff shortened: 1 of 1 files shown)


charsmith:
For greater context, I was playing with this and these tests were failing:

tests.test_execute.TestBrokenNotebook

```
  File "/data/code/papermill/tests/test_execute.py", line 39, in test
    execute_notebook(path, result_path)
  File "/data/code/papermill/papermill/execute.py", line 114, in execute_notebook
    nb.metadata.papermill['exception'] = any([cell.metadata.papermill.exception for cell in nb.cells])
AttributeError: 'dict' object has no attribute 'exception'
```

And 

tests.test_execute.TestNotebookHelpers

```
  File "/data/code/papermill/tests/test_execute.py", line 22, in test
    execute_notebook(nb_test1_fname, nb_test1_executed_fname, {'msg': 'Hello'})
  File "/data/code/papermill/papermill/execute.py", line 114, in execute_notebook
    nb.metadata.papermill['exception'] = any([cell.metadata.papermill.exception for cell in nb.cells])
AttributeError: 'dict' object has no attribute 'exception'
```

In the debugging output these messages occurred:

```
traitlets: ERROR: Notebook JSON is invalid: '*' is not of type u'integer', u'null'
```

And as a note, if you open a running notebook with Jupyter while papermill is running, it gives this error:

```
Notebook validation failed: '*' is not of type 'integer', 'null':
"*"
```

rgbkrk:
Thanks @charsmith! We should make a patch release of this too.

mpacer:
This still sets the execution count to `"*"` which is a violation of the format. If the execution is interrupted without resetting the cell's `execution_count` it will produce an invalid notebook file. It also is why you're getting those last two errors (in debugging and when opening the notebook)… 

If you don't know the execution count number yet, there is no way to communicate that to the on-disk representation other than to say it is `None`

I'll make a PR to fix this…

PR #28, opened 2017-08-29
nb cell errors will result in raising a PapermillExecutionError

ewmassey:
A notebook cell execution error will result in the notebook saving to disk as well as PapermillExcecutionError being raised. The PapermillExecutionError will output both it's own traceback as well as the traceback for the cell that had the exception.

### DIFF
--- a/papermill/exceptions.py
+++ b/papermill/exceptions.py
@@ -20,7 +20,7 @@ class PapermillExecutionError(PapermillException):
         self.evalue = evalue
         self.traceback = traceback
         message = "\n" + 75 * "-" + "\n"
-        message += "Exception encountered at Out [%s]:\n" % str(exec_count)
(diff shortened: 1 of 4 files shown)


papermill/exceptions.py
```diff
@@ -9,3 +9,20 @@ class FileExistsError(AwsError):
...
+
+class PapermillExecutionError(PapermillException):
+    """Raised when an exception is encountered in a notebook."""
+    def __init__(self, exec_count, source, ename, evalue, traceback):
+        self.exec_count = exec_count
+        self.source = source
+        self.ename = ename
+        self.evalue = evalue
+        self.traceback = traceback
+        message = "\n" + 75 * "-" + "\n"
```
rgbkrk: This is a nice little pass up of the error in the user's code.

PR #31, opened 2017-09-05
change execution count to be set to None not *

mpacer:
addresses issues left open by #27 regarding the setting of the execution count to `"*"`.

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -58,7 +58,7 @@ def preprocess(self, nb, resources):
     with futures.ThreadPoolExecutor(max_workers=1) as executor:
         for index, cell in enumerate(nb.cells):
             if hasattr(cell, "execution_count"):
-                cell.execution_count = "*"
+                cell.execution_count = None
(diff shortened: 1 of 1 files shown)


ewmassey:
For context, the reason we want to set the execute count to "*" is so papermill can inform the user which cell it is executing when it is executing the notebook. For long running jobs users may want to see the progress of their notebook and having the "*" show up in the output gives the user a similar experience to running the notebook in browser. 

Given all the issues that are being encountered due to this being an invalid format we can find another way to inform the user which cell is being executed.

ewmassey:
I would like to close this PR in favor of this one https://github.com/nteract/papermill/pull/32. @mpacer I'll let you close it since you created it.

ewmassey:
@mpacer thanks for doing this. I went ahead and pulled this commit into #32. The approach in #32 will include the status of papermill in the metadata instead of violating the spec like we're doing now.

rgbkrk:
Thanks @mpacer!

PR #32, opened 2017-09-06
Removing * and replacing with papermill field 'status'

ewmassey:
Setting the execution count to "\*" is a violation of the ipynb format and causing bugs in papermill and when trying to open partially complete notebooks in jupyter. The goal of using "\*" was so users could open running notebooks and see what cell papermill was currently running. This PR moves away from the "\*" execution count value in favor of updating a status field in the papermill metadata (i.e. node.metadata.papermill.status). We can then choose down the road how to use the status field for tools like commuter to indicate which cell is being run without violating the notebook spec.

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -58,7 +58,7 @@ def preprocess(self, nb, resources):
     with futures.ThreadPoolExecutor(max_workers=1) as executor:
         for index, cell in enumerate(nb.cells):
             if hasattr(cell, "execution_count"):
-                cell.execution_count = "*"
+                cell.execution_count = None
             future = executor.submit(write_ipynb, nb, output_path)
             t0 = datetime.datetime.utcnow()
             try:
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -14,6 +14,10 @@ from papermill.iorw import load_notebook_node, write_ipynb, read_yaml_file
 
 from six import string_types
 
+PENDING = "pending"
+RUNNING = "running"
+COMPLETED = "completed"
+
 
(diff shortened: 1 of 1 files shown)


papermill/execute.py
```diff
@@ -41,24 +45,24 @@ def preprocess(self, nb, resources):
...
             cell.execution_count = None
 
         # Clear out the papermill metadata for each cell.
         cell.metadata['papermill'] = dict(
             exception=None,
             start_time=None,
             end_time=None,
-            duration=None
+            duration=None,
+            status=PENDING  # pending, running, completed
```
rgbkrk: We definitely need to start speccing the metadata more formally (JSON Schema). We can get this metadata acknowledged in commuter and nteract pretty easily as well. It would help if the metadata had a version too, for our own sake. It's also good to think about how we'll upconvert from old versions of the papermill metadata to new versions. In this case it's new fields, just something to keep in mind for stability and maintenance sanity.

papermill/execute.py
```diff
@@ -41,24 +45,24 @@ def preprocess(self, nb, resources):
     # Reset the notebook.
     for cell in nb.cells:
```
mpacer: It seems like resetting the notebook in a papermill specific manner should be its isolated function. That way it's easier to unit test and this kind of metadata manipulation should definitely be unit tested.

mpacer: It seems like resetting the notebook in a papermill specific manner should be its isolated function. That way it's easier to unit test and this kind of metadata manipulation should definitely be unit tested.

PR #35, opened 2017-09-19
Error message at top of notebook if exception encountered.

ewmassey:
Many users are confused when their notebook fails as to whether it is an issue or with papermill. This "banner" will be rendered whenever there is a notebook generated exception.

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -122,7 +122,7 @@ def execute_notebook(notebook, output, parameters=None, kernel_name=None):
 
     # Write final Notebook to disk.
     write_ipynb(nb, output)
-    raise_for_execution_errors(nb)
+    raise_for_execution_errors(nb, output)
 
 
 def _parameterize_notebook(nb, kernel_name, parameters):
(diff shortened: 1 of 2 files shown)


papermill/execute.py
```diff
@@ -231,5 +238,11 @@ def raise_for_execution_errors(nb):
                     traceback=output.traceback
                 )
             break
+
     if error:
+        # Write notebook back out with the Error Message at the top of the Notebook.
+        error_msg = ERROR_MESSAGE_TEMPLATE % str(error.exec_count)
+        error_msg_cell = nbformat.v4.new_markdown_cell(source=error_msg)
+        nb.cells = [error_msg_cell] + nb.cells
+        write_ipynb(nb, output_path)
```
rgbkrk: This is an error from their execution or from papermill? I'm confused.

ewmassey: This is an error encountered by the running notebook. I can update the var name to make that more clear if you'd like.

rgbkrk: What I'd probably write (in the template above) is:

```
This notebook encountered an exception at `In [%s]` while being run by papermill.
```

rgbkrk: What I'm a bit lost on is why we're not embedding the classic error / stacktrace within the notebook where it errored (or perhaps it does that anyways...).

PR #38, opened 2017-09-24
Simplify naming of cli options

willingc:
- simplify naming (parameters -> param)
- make long and short name of options consistent with leading letter (`--file_param`, `-f`)
- add help options
- change to `src` and `dst` since greater clarity than `notebook` and `output`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -66,8 +66,8 @@ Execute via the Python API
    import papermill as pm
 
    pm.execute_notebook(
-      src = 'path/to/input.ipynb',
-      dst = 'path/to/output.ipynb',
+      'path/to/input.ipynb',
+      'path/to/output.ipynb',
       parameters = dict(alpha=0.6, ratio=0.1)
    )
 
--- a/README.rst
+++ b/README.rst
@@ -66,8 +66,8 @@ Execute via the Python API
    import papermill as pm
 
    pm.execute_notebook(
-      notebook_path = 'path/to/input.ipynb',
-      output_path = 'path/to/output.ipynb',
+      source = 'path/to/input.ipynb',
+      dest = 'path/to/output.ipynb',
       parameters = dict(alpha=0.6, ratio=0.1)
    )
 
@@ -82,7 +82,7 @@ Amazon S3 account:
    $ papermill local/input.ipynb s3://bkt/output.ipynb -p alpha 0.6 -p l1_ratio 0.1
 
 
-.. _PythonBinding: 
+.. _PythonBinding:
 
 Python In-notebook Bindings
 ---------------------------
--- a/README.rst
+++ b/README.rst
@@ -66,8 +66,8 @@ Execute via the Python API
    import papermill as pm
 
    pm.execute_notebook(
-      source = 'path/to/input.ipynb',
-      dest = 'path/to/output.ipynb',
+      src = 'path/to/input.ipynb',
+      dst = 'path/to/output.ipynb',
       parameters = dict(alpha=0.6, ratio=0.1)
    )
 
--- a/papermill/cli.py
+++ b/papermill/cli.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-
 """Main `papermill` interface."""
 
 import base64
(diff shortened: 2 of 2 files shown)


README.rst
```diff
@@ -66,8 +66,8 @@ Execute via the Python API
    import papermill as pm
 
    pm.execute_notebook(
-      notebook_path = 'path/to/input.ipynb',
-      output_path = 'path/to/output.ipynb',
+      src = 'path/to/input.ipynb',
+      dst = 'path/to/output.ipynb',
```
rgbkrk: I love this naming. For backwards compatibility though, we would need to make `notebook_path` and `output_path` be deprecated and pass it through on the new `src` and `dst` keyword args.

papermill/cli.py
```diff
@@ -1,38 +1,66 @@
+# -*- coding: utf-8 -*-
+
+"""Main `papermill` interface."""
```
rgbkrk: Docstrings woo!

papermill/cli.py
```diff
@@ -1,38 +1,66 @@
...
-@click.argument('output')
-@click.option('--parameters', '-p', help="Parameters to pass to the parameters cell.", multiple=True, nargs=2)
-@click.option('--raw_parameters', '-r', help="Parameters to be read as raw string.", multiple=True, nargs=2)
-@click.option('--parameters_file', '-f', help="Path to YAML file containing parameters.")
-@click.option('--parameters_yaml', '-y', help="YAML string to be used as parameters.")
-@click.option('--parameters_base64', '-b', help="Base64 encoded YAML string as parameters.")
-@click.option('--kernel', '-k', help="Kernel name to run the notebook with.")
-def papermill(notebook, output, parameters, raw_parameters, parameters_file, parameters_yaml, parameters_base64,
-              kernel):
+@click.command(context_settings=dict(help_option_names=['-h', '--help']))
```
rgbkrk: Neat, I didn't know about this.

papermill/cli.py
```diff
@@ -1,38 +1,66 @@
...
+@click.argument('src')
+@click.argument('dst')
+@click.option(
+    '--param', '-p',
```
rgbkrk: If there's a way to do it, we should definitely handle the old version of this `--parameters` too.

papermill/cli.py
```diff
@@ -1,38 +1,66 @@
...
+    help='Parameters to pass to the parameters cell.',
+    multiple=True, nargs=2
+)
+@click.option(
+    '--raw_param', '-r',
+    help='Parameters to be read as raw string.',
+    multiple=True, nargs=2
+)
+@click.option(
+    '--file_param', '-f',
```
rgbkrk: What would you think of `--params_via_file` or `--param_file`? It's really hard for me to read `file_param` and not think of it as "the file param" which I would interpret as being a redundant version of a "file" param.

README.rst
rgbkrk: I just noticed that this needs to go one of two ways here, because no changes were made to `execute_notebook`'s function:

Either we'll need to change `execute_notebook` or this example should go without named arguments (for now).

rgbkrk:
I gave some feedback which is mostly a hesitation to change an interface we currently ship/support. Improving the CLI interface is welcome here, I think it's a step in the right direction.

Now that I've read your description through again:

> make long and short name of options consistent with leading letter

ohhhhhh that's why, I get it now 🤔 

> change to src and dst since greater clarity than notebook and output

This is definitely a great idea.

> simplify naming (parameters -> param)

I'm on the fence about this one, since there _is_ a shorthand and the long form versions make it explicit and verbose.

README.rst
willingc: I think for this one that unnamed arguments are the way to go esp since the files are name input and output. Agree on the deprecation and pass through. (Wasn't sure how many folks were using.)

rgbkrk: At least for my friendly Netflix colleagues, it's in the dozens. We can always make this a short deprecation cycle, I just like to not cause too much thrashing for people. I did talk to Eric about making the papermill interface allow publishing too, which would also change this fundamentally too.

rgbkrk:
Since you made me curious, I ran BigQuery with the following to find out the number of downloads:

```
SELECT COUNT(*) as download_count
FROM TABLE_DATE_RANGE(
  [the-psf:pypi.downloads],
  TIMESTAMP("2015-05-01"),
  CURRENT_TIMESTAMP()
)
WHERE file.project="papermill"
```

Result is... 55,976

Which is a bit high. I know that at Netflix we install it in our base image but we don't install it _that_ many times. Wish I had a decent way to normalize against mirroring data.

rgbkrk:
Oh wait, I have a perfectly useless package I can use for normalizing against mirroring data (or maybe not since we ship `papermill` frequently).:

```
SELECT COUNT(*) as download_count
FROM TABLE_DATE_RANGE(
  [the-psf:pypi.downloads],
  TIMESTAMP("2015-05-01"),
  CURRENT_TIMESTAMP()
)
WHERE file.project="kyle"
```

Result: 3,801

README.rst
willingc: Let's do this. I'm going to leave the interface as is for now. When we handle publishing, we can extend/deprecate/etc. the interface. I changed the naming before I updated the formatting and the format change got rid of most of the long name yuckiness.

willingc:
screenshot 2017-09-24 14 25 18

:smile:

papermill/cli.py
```diff
@@ -1,22 +1,53 @@
...
-@click.option('--parameters_file', '-f', help="Path to YAML file containing parameters.")
-@click.option('--parameters_yaml', '-y', help="YAML string to be used as parameters.")
-@click.option('--parameters_base64', '-b', help="Base64 encoded YAML string as parameters.")
-@click.option('--kernel', '-k', help="Kernel name to run the notebook with.")
-def papermill(notebook, output, parameters, raw_parameters, parameters_file, parameters_yaml, parameters_base64,
-              kernel):
-    """Utility for executing a single notebook on a container."""
+@click.command(context_settings=dict(help_option_names=['-h', '--help']))
+@click.argument('notebook_path')
+@click.argument('output_path')
```
rgbkrk: I think you can make these `src` and `dst` for the sake of the command line since they're positional arguments, then...

papermill/cli.py
```diff
@@ -1,22 +1,53 @@
...
+)
+@click.option(
+    '--parameters_base64', '-b',
+    help='Base64 encoded YAML string as parameters.'
+)
+@click.option(
+    '--kernel', '-k',
+    help='Name of kernel to run.'
+)
+def papermill(notebook_path, output_path, parameters, parameters_raw,
```
rgbkrk: ...set these to `src, dst,`...

papermill/cli.py
```diff
@@ -28,11 +59,10 @@ def papermill(notebook, output, parameters, raw_parameters, parameters_file, par
         parameters_final = {}
         for name, value in parameters:
             parameters_final[name] = _resolve_type(value)
-        for name, value in raw_parameters:
+        for name, value in parameters_raw:
             parameters_final[name] = value
 
-    # Notebook Execution
-    execute_notebook(notebook, output, parameters_final, kernel_name=kernel)
+    execute_notebook(notebook_path, output_path, parameters_final, kernel_name=kernel)
```
rgbkrk: ...and pass `src`, `dst` in directly here as `execute_notebook(src, dst, parameters_final, kernel_name=kernel)` or even as `execute_notebook(notebook_path=src, output_path=dst, parameters_final, kernel_name=kernel)`

rgbkrk:
Going ahead and merging, thanks for the wordsmithing and clarity here. 😄

papermill/cli.py
rgbkrk: I went ahead and merged, you're welcome to make these changes in another PR. I wanted to make sure we brought in what you've already done.

PR #41, opened 2017-09-27
Find "error" outputs beyond the first output element of a cell.

adparker:
Ran into an issue where a notebook exception was being missed (and not raised) because the output element containing the exception was not the first output element of the cell. I also added a new test case.


papermill/execute.py
```diff
@@ -237,7 +237,7 @@ def raise_for_execution_errors(nb, output_path):
                     evalue=output.evalue,
                     traceback=output.traceback
                 )
-            break
+                break
```
rgbkrk: Classic indents

PR #42, opened 2017-09-28
using regex to remove ansi escape for colors

ewmassey:
This removes the color escape chars so that it doesn't muddy up the exception outputs.

### DIFF
--- a/papermill/exceptions.py
+++ b/papermill/exceptions.py
@@ -1,3 +1,5 @@
+import re
+
 
 class AwsError(Exception):
     """Raised when an AWS Exception is encountered."""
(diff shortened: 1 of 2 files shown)


rgbkrk:
This is only for while papermill is running, right? Not for inside the notebook where we can have all the ansi codes we like?

adparker:
Thanks for the quick turn around!

papermill/exceptions.py
```diff
@@ -21,8 +23,11 @@ def __init__(self, exec_count, source, ename, evalue, traceback):
         self.traceback = traceback
         message = "\n" + 75 * "-" + "\n"
         message += 'Exception encountered at "In [%s]":\n' % str(exec_count)
+        tb = ""
         for line in traceback:
-            message += line + "\n"
+            tb += line + "\n"
+        ansi_escape = re.compile(r'\x1b[^m]*m')
```
rgbkrk: The complete regex you probably want is `r'\x1b\[[\d;]+m`. However, there is a module called `ansiwrap` with a function `strip_color`.

rgbkrk:
:shipit:

PR #43, opened 2017-10-02
added progress bar

ewmassey:
Adding progress bar via tqdm. Could use some eyes on how I pass the flag.


SynapticSage:
Would it be useful to let users switch tqdm() to the tqdm_notebook() function? It works exactly the same, but prints progress bars better in notebooks, without forcing new progress messages onto a new line. Perhaps an interface could be exposed for users to set that function instance?

MSeal:
@SynapticSage Could you make a new issue for the idea? It's likely people won't look here on the merged pull request thread. Thanks :)

Issue #46, opened 2017-10-06
Get code coverage to ~95%


rgbkrk:
We can do it and anyone can help. 😄 

I've tagged this as Hacktoberfest friendly, feel free to ask questions.

## Requirements

Must have knowledge:

* Python

Optional (you can learn it):

* [py.test](https://docs.pytest.org/en/latest/)

## Issue Creation

It's ok if people want to create issues for each of the untested areas then go ahead and tackle them

## TODO

* [ ] List how to generate a coverage report for papermill

achou11:
Would love to attempt this help out with this if I can! I don't have any experience with testing or code coverage but would like to learn.

rgbkrk:
You're on the hook for some example notebooks in `vdom` too! :wink:

To get started, I recommend cloning this repository and looking at the current tests in `papermill/tests`. You can see how the tests run by looking at Travis' configuration here: https://github.com/nteract/papermill/blob/master/.travis.yml#L17

Once you've installed `pytest` yourself (`pip install -U pytest`), you should also be able to run the test suite locally on your clone of this repository. If you run into snags, let us know.

achou11:
Ok thanks! A couple questions before I dive into it: 

- Do I have to worry about what Python version I'm using? I've been running into some problems with 2.7 and I usually stick to 3 anyways. 

- Should I have a virtual environment set up? I already have all of the dependencies installed so I didn't see the need to do this, but it seems like good practice. 

- Am I focusing on writing tests or fixing to match tests? Or does it depend on the situation? Seems like a silly question but like I said, I don't really have much experience with this aspect 😅 (feel free to point me to any resources that would answer this)

rgbkrk:
Feel free to stick with Python 3, you can always let Travis do the checking to make sure it works on both Python 2 and Python 3.

> Should I have a virtual environment set up? I already have all of the dependencies installed so I didn't see the need to do this, but it seems like good practice.

It helps so you insulate yourself from changes (even your own). It's not a strict requirement though. I'm running `papermill` from master on my main python install.

> Am I focusing on writing tests or fixing to match tests?

For this, you'll be writing tests, in a similar style to the current ones. If you look [at papermill's codecov](https://codecov.io/gh/nteract/papermill/tree/master/papermill) you can see which files have the least coverage, then click in to files. I'd start off with [`api.py`](https://codecov.io/gh/nteract/papermill/src/master/papermill/api.py) and see if there's something you could exercise in a few lines of python.

[...]

PR #47, opened 2017-10-07
omit the versioneer script

rgbkrk:
The versioneer script comes from [python-versioneer](https://github.com/warner/python-versioneer) which is tested in its own repo. We don't want it to be part of our code coverage.


rgbkrk:
I'm going to go ahead and merge this to make it easier for newcomers to go after #46.

PR #49, opened 2017-10-08
Add tests for notebook_dir, rewrote float and int tests

aaronmak:


papermill/cli.py
```diff
@@ -97,19 +97,17 @@ def _resolve_type(value):
...
 
 def _is_float(value):
-    try:
-        float(value)
-    except ValueError:
+    if type(value) is float:
+        return True
+    else:
         return False
-    return value
```
rgbkrk: 👍

rgbkrk:
/cc @ewmassey ^^ some semantics changed above, they make way more sense to me than the old `_is_float` code...

Oh wait... it just dawned on me... It's not about the type explicit checking, this is about coercion from strings for the cli.

papermill/cli.py
```diff
@@ -97,19 +97,17 @@ def _resolve_type(value):
 
 
 def _is_int(value):
-    try:
-        int(value)
-    except ValueError:
+    if type(value) is int:
+        return True
+    else:
```
rgbkrk: Welp, I made a mistake. The reason this used `int` to see if a value was an int was that it was converting string parameters from the command line (since it only passes strings). The casting was a way to convert from string to int, float, etc.

I can either revert this _or_ you can make a new PR that puts this back in place with a comment about using casting for conversion...

papermill/tests/test_cli.py
```diff
@@ -22,11 +22,11 @@ def test_resolve_type(test_input, expected):
 
 
 @pytest.mark.parametrize("value,expected", [
-    (13.71, 13.71),
+    (13.71, True),
```
rgbkrk: We should probably make a test set here that is `("2.3", True)`

papermill/cli.py
aaronmak: Ohh, I see what you mean. No worries, I'll put it back in place and add a comment

aaronmak: So just to be clear, is it okay even if an int was converted to a float? Based on the old version, in the `_resolve_type` function, any int will be converted to a float and it will never reach the `_is_int` cast.

rgbkrk: Yeah that's fine, float should take primacy.

PR #52, opened 2017-10-17
metrics method for accessing metrics collected during execution

ewmassey:


rgbkrk:
Awesome, looks like a good way to expose the metrics more easily.

Issue #53, opened 2017-10-25
`record`ing non-JSON'able objects


betatim:
I tried to `record` a numpy array but that fails somewhere in a step to clean the JSON. I think the problem is that there is no builtin way to convert a numpy array to JSON.

What is the recommendation for outputting numpy arrays? For example I could imagine that "arrays tend to be too big, so store them somewhere else and `record` the path to it" is one option. An alternative would be to add a step to `pickle`/serialize types that can't be stored as JSON in `record` before dumping them into the notebook.

I'd be up for implementing the latter option but wanted to discuss ideas first.

rgbkrk:
Great question!

> "arrays tend to be too big, so store them somewhere else and `record` the path to it"

That would be my recommendation and I'd like to make it a simple built-in if it makes sense across users. I think it requires some infrastructure as well as opinion. The stance I'd likely take inside Netflix as a paved path would be to `record` the array (or pandas dataframe) likely using [Arrow to store it to HDFS](http://arrow.apache.org/docs/python/filesystems.html) or S3. Then the field we store in a papermill record is likely the full path to the record (which can be an S3 path, or an HDFS path). What object storage would you end up using?

How lossy would it be for you to use Pandas' JSON Table Schema + row oriented JSON option?

> An alternative would be to add a step to pickle/serialize types that can't be stored as JSON in record before dumping them into the notebook.

As much as possible we want to keep the data types language agnostic (as there are [R bindings for papermill](https://github.com/nteract/papermillr) as well). I think it would be ok to serialize them, so long as you're the one expecting it on the other side of reading them (where you == you and your collaborators).

rgbkrk:
@charsmith -- any thoughts here?

mpacer:
If you are going the pickle route, it's probably better to use [dill](https://github.com/uqfoundation/dill).

rgbkrk:
Shameless plug - have you tried [`cloudpickle`](https://github.com/cloudpipe/cloudpickle)?

lukasheinrich:
as mentioned in my PR https://github.com/nteract/papermill/pull/56 there is some usecase of persisting the data that is `record`ed.. right not it assumes it's jsonable via panda's `to_json()` but perhaps other persistency approaches could be made pluggable

betatim:
I think `pickle` (and friends) aren't the right way to go because you'd want to be able to load the data in R (and maybe other languages).

[...]

PR #56, opened 2017-10-25
execution: add option to write notebook dataframe to disk as JSON

lukasheinrich:
Hi,

papermill allows the notebook to attach arbitrary data during it's execution, it would be useful to be able to write that file to disk, so that it can be further manipulated by tools outside of the papermill ecosystem / outside of python even.

A similar option could be useful for the `.data` property (which seems to be a JSON'able dict

* adds an option in execute_notebook() to write dataframe to disk via pandas to_json() function
* adds corresponding CLI option --dataframe-file

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -14,6 +14,7 @@ from tqdm import tqdm
 from papermill.conf import settings
 from papermill.exceptions import PapermillException, PapermillExecutionError
 from papermill.iorw import load_notebook_node, write_ipynb, read_yaml_file, get_pretty_path
+from papermill.api import read_notebook
 
 PENDING = "pending"
 RUNNING = "running"
@@ -125,7 +126,7 @@ def log_outputs(cell):
 Preprocessor.preprocess = preprocess
 
 
-def execute_notebook(notebook, output, parameters=None, kernel_name=None, progress_bar=True, log_output=False):
+def execute_notebook(notebook, output, parameters=None, kernel_name=None, progress_bar=True, log_output=False, dataframe_file=None):
     """Executes a single notebook locally.
 
     Args:
@@ -169,7 +170,10 @@ def execute_notebook(notebook, output, parameters=None, kernel_name=None, progre
     # Write final Notebook to disk.
     write_ipynb(nb, output)
     raise_for_execution_errors(nb, output)
-
+    if dataframe_file:
+        outputnb = read_notebook(output)
+        with open(dataframe_file,'w') as dfile:
+            dfile.write(outputnb.dataframe.to_json())
 
 def _parameterize_notebook(nb, kernel_name, parameters):
 
--- a/papermill/cli.py
+++ b/papermill/cli.py
@@ -47,9 +47,13 @@ from papermill.iorw import read_yaml_file
     '--log-output/--no-log-output', default=False,
     help="Flag for writing notebook output to stderr."
 )
+@click.option(
+    '--dataframe-file','-d', default=False,
+    help="JSON representation of output dataframe"
+)
 def papermill(notebook_path, output_path, parameters, parameters_raw,
               parameters_file, parameters_yaml, parameters_base64, kernel,
-              progress_bar, log_output):
+              progress_bar, log_output, dataframe_file):
     """Utility for executing a single notebook on a container.
 
     Take a source notebook, apply parameters to the source notebook,
@@ -77,7 +81,8 @@ def papermill(notebook_path, output_path, parameters, parameters_raw,
         parameters_final,
         kernel_name=kernel,
         progress_bar=progress_bar,
-        log_output=log_output
+        log_output=log_output,
+        dataframe_file=dataframe_file
     )
 
 


papermill/execute.py
```diff
@@ -169,7 +170,10 @@ def execute_notebook(notebook, output, parameters=None, kernel_name=None, progre
     # Write final Notebook to disk.
     write_ipynb(nb, output)
     raise_for_execution_errors(nb, output)
-
+    if dataframe_file:
+        outputnb = read_notebook(output)
```
lukasheinrich: i'm not sure this is the most elegant way, but `nb.dataframe` didn't seem to work

betatim:
What is the advantage of writing it to an external file instead of the output notebook (which is just a big JSON document)?

lukasheinrich:
in that case you need to understand the notebook format. Say if you just want the raw `record`ed data (actually this may be more applicable `.data` instead of `.dataframe`), the downstream tool should not need to be able to know that it came from a notebook at all.

papermill/execute.py
rgbkrk: Yeah it looks like it's not easy to get a `papermill.api.Notebook` object straightaway with this.

Looks like you'd need to take this `nb` (nbformat object) and do the following:

```python
from .api import Notebook

n = Notebook() # we should probably let this take an `nbformat` node object
n.path = 'whateverthepathis'
n.node = nb
```

rgbkrk:
First off, thanks for making a contribution and diving straight in to what's here!

Approach wise, we'd like to keep the container format be the notebook itself while writing functions within papermill (whether these Python bindings, R bindings, or a new binding) that know how to extract the right records from the notebook.

My worry over writing out a separate JSON file is that there's more than one way to serialize a dataframe to JSON. Once we start doing it, we have to support it.

papermill/execute.py
rgbkrk: Probably in a separate PR, I think we should let `papermill.api.Notebook` accept both the `path` (optional) and the `node` in it's `__init__` to be able to take the `nbformat` node. That would be super helpful for this PR and future ones.

lukasheinrich:
I agree multiple versions to persist may be an issue, but that could actually be supported in a uniform way with e.g. https://github.com/nteract/papermill/issues/53

i.e. there might be "persistence options" in general that could include
* format (json, pickle, dill, cloudpickle..)
* format-specific options (between e.g. `.dataframe`/`.data` and output format) . if dataframe and if format is json, what what orientation

I guess one way to think of the disk persistence is a very low-level POSIX bindings. The use-case I'm looking at is e.g. generating some data and processing it further with `jq` or other tools that natively support JSON structures.

Alternatively, the notebook could of course write that data to an actual json file itself, but it seems the record's were purposefully added to the notebook to give the notebook some "output" semantics (to match the "input" which are the parameters)

lukasheinrich:
by the way what's the model for `.data` is this expected to be a JSON'able thing?

papermill/execute.py
lukasheinrich: something like this https://github.com/nteract/papermill/pull/57 ?

rgbkrk: Merged, thanks!

rgbkrk:
I'm going to have to look through it again and come back to this PR after thinking about it overnight. The primary lead on this library just had a baby, so I'll wait until he gets back from paternity leave to pester him.

Hey @adparker what do you think of this PR + discussion?

lukasheinrich:
Hi, should we revisit this PR?

papermill/execute.py
betatim: Coming back to this now I am confused why `nb == outputnb` isn't true. Would be good if we can add a comment because I predict we will be confused by this again :-/

rgbkrk:
[My opinion on writing out sidecar data files](https://github.com/nteract/papermill/pull/56#issuecomment-339469364) still holds. However, I don't wish to be a blocker.

What do you think @mseal @willingc?

MSeal:
I think I agree with your opinion @rgbkrk. In principle the capability sounds like it makes certain workflows simpler. In practice at a general level this would open up a rather large shoot-your-self-in-the-foot door  where giant dataframes or dataframes generated from a variety of sources would need to be supported. A few concerns come to mind: custom types, pandas version support/new features, data size, graceful failures, and testing potential downstream consumers.

I think the idea could be useful in more controlled environments but I don't think it belongs in papermill as a generally supported feature.

willingc:
@rgbkrk For now, let's stick with simple. Until we have this fully documented and coverage increased, I'm hesitant to add an additional file to keep track of. Once that happens, let's revisit which approach makes most sense longer-term.

rgbkrk:
> Until we have this fully documented and coverage increased, I'm hesitant to add an additional file to keep track of. Once that happens, let's revisit which approach makes most sense longer-term.

Based on this, let's go ahead and close for now @lukasheinrich. Once we've got tests and documentation up, we can start entertaining ideas like this. I'm still super appreciative that you went through the review and followed up with some smaller PRs to fix up other bits in papermill. Keep it coming!

PR #58, opened 2017-10-26
[MRG] Swap order of test for `int` and `float`

betatim:
Every int is also a float, so checking for float first means we never
create any integers.

### DIFF
--- a/papermill/cli.py
+++ b/papermill/cli.py
@@ -88,10 +88,10 @@ def _resolve_type(value):
         return False
     elif value == "None":
         return None
-    elif _is_float(value):
-        return float(value)
     elif _is_int(value):
         return int(value)
+    elif _is_float(value):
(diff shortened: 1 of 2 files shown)


rgbkrk:
Won't this downcast all ints to floats though?

Nevermind, I'm wrong.

```
>>> int('4.3')
ValueError: invalid literal for int() with base 10: '4.3'
```

betatim:
c7b978c changes an existing test, so we should discuss it. IMHO `_resolve_type` is meant to be used on stuff we parsed from the command line arguments. In python these will always be strings (or bytes). This means we do not need to do "the right thing" when passing in an `int` or `float`. Hence removing the test for that situation.

If we do want to support that then we need to make `_resolve_type` a bit smarter by checking the type of its argument and doing different things depending on that type.

PR #61, opened 2017-10-31
rely on __future__'s unicode_literals

rgbkrk:
We ran into an issue where a user had some `!pip install` that had unicode output that wasn't being encoded and decoded properly. This switches the section of code that mixed string concatenation between Python 2 strings and unicode strings over to only using unicode strings. While I was at it I brought in `unicode_literals` from `__future__`.

/cc @mpacer @charsmith

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -1,3 +1,5 @@
+ # -*- coding: utf-8 -*-
+
 from __future__ import unicode_literals
 
 import datetime
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 import datetime
 import os
 import sys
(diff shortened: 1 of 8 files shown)


mpacer:
I'd also add  `# -*- coding: utf-8 -*-` to the beginning of most of those files in case the issue has to do with parsing the literal string sources when the underlying system encoding may not be utf-8.

See https://stackoverflow.com/questions/4872007/where-does-this-come-from-coding-utf-8 for a quick explanation
See https://www.python.org/dev/peps/pep-0263/ for a more comprehensive look into why this is needed.

rgbkrk:
Ok cool, thanks. Feel free to push to this branch. We could probably add another test that exercises this as well.

mpacer:
I'd definitely push if I were able to reproduce the error locally :/ so I can know that this makes a difference.

papermill/execute.py
```diff
@@ -106,26 +112,31 @@ def log_outputs(cell):
     for output in cell.get("outputs", []):
         if output.output_type == "stream":
             if output.name == "stdout":
-                stdouts.append("".join(output.text))
+                stdouts.append(u"".join(output.text))
```
mpacer: These shouldn't be necessary anymore.

rgbkrk: Yeah since we imported Unicode literals from future

PR #62, opened 2017-11-01
Add test for display function

menglewis:
This adds 2 tests: One for the `display` function and another for initializing a Notebook with a path but no node specified.


ewmassey:
Thanks for doing this @menglewis.

PR #63, opened 2017-12-06
only install futures for python 2

charsmith:
Unfortunately we are dealing with a pypi repo that tries to install futures==3.2.0 in python3.  Since we don't need futures in python3, let's just filter.


rgbkrk:
Shipping now as 0.10.1

PR #71, opened 2017-12-19
Changed import paths to be relative; rather than a mix of relative and absolute

MSeal:
Relative imports force everything to load locally and protects from accidental bad import patterns.


rgbkrk:
Just stated in person, this'll need a rebase since I merged @harph's PR. 😄

PR #73, opened 2017-12-19
refactor source

willingc:
- clean up parameter passing of tuples in `api.py`
- rename error prone `dfs` and `nbs` to more meaningful collection names
- remove extraneous line in `setup.py`\
- clean up style


charsmith:
👍

Issue #68, opened 2017-12-19
Append instead of replace parameter cell


MSeal:
Cells overwrite the contents of the parameter tagged cell which is non-intuitive and easy to mess up. Instead it would be more reasonable if the tagged cell used values there as default values and simply appended its contents to the cell definition. Repeated names would still be overwritten and non-papermill execution could enumerate sane values without breaking papermill execution as it is today. But, it would allow for easier assignment of defaults without adding a new cell. Today all the notebooks I've seen for papermill leave a blank cell for parameters which has to be explained to each person who sees a papermill notebook as opposed to a normal notebook when there shouldn't be so much of a difference.

rgbkrk:
Yeah, I would definitely love if the values in the cells were treated as default values and left if not overridden.

betatim:
(related #54)

Issue #67, opened 2017-12-19
Default to first cell or prefix a new cell for parameters


MSeal:
When a user doesn't specify a parameter tag cell it would be nice if papermill defaulted to some sane logical setting. Nominally parameterizing the beginning of the notebook seems reasonable and would make adoption of existing notebooks quicker when they have naive inputs.

PR #74, opened 2017-12-19
Changed parameters to append cells instead of replace

MSeal:
Also allowed for notebooks with no parameter tags

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -141,7 +141,7 @@ def execute_notebook(notebook,
 
     Args:
         notebook (str): Path to input notebook.
-        output (str): Path to save exexuted notebook.
+        output (str): Path to save executed notebook.
         parameters (dict): Arbitrary keyword arguments to pass to the notebook parameters.
         kernel_name (str): Name of kernel to execute the notebook against.
         progress_bar (bool): Flag for whether or not to show the progress bar.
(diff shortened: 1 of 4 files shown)


MSeal:
Solves https://github.com/nteract/papermill/issues/68 and https://github.com/nteract/papermill/issues/67

papermill/execute.py
```diff
@@ -195,11 +195,15 @@ def _parameterize_notebook(nb, kernel_name, parameters):
...
     # Remove the old cell and replace it with a new one containing parameter content.
     param_cell_index = _find_parameters_index(nb)
-    old_parameters = nb.cells[param_cell_index]
-    before = nb.cells[:param_cell_index]
+    if param_cell_index >= 0:
+        old_parameters = nb.cells[param_cell_index]
+    else:
+        old_parameters = nbformat.v4.new_code_cell() # Fake cell
+        old_parameters.metadata['tags'] = ['parameters']
+    before = nb.cells[:param_cell_index + 1]
```
harph: The current file on master has `before = nb.cells[:param_cell_index]`. Why are you adding `+1` to the slice? Wouldn't this include the param cell as well?

MSeal: Yes it does, that's the intention now. Insert instead of replace. This the cell you tagged is treated as prefix defaults to input parameters

MSeal: Input
input
Output
output

harph: got it

kainoa21: This is great.  Do you think it makes sense to update the tag(s) of the output cells to reflect that one is the original cell and one is the newly appended cell with the passed parameters?

MSeal: @kainoa21 That sounds like a good idea for readability, do you want to make an issue for it as a reminder to make that improvement?

kainoa21: https://github.com/nteract/papermill/issues/84

PR #76, opened 2017-12-19
Added unittest for HttpHandler

harph:


papermill/tests/test_iorw.py
```diff
@@ -1 +1,53 @@
-import pytest
+import json
+import unittest
+import six
+if six.PY3:
+    from unittest.mock import Mock, patch
+else:
+    from mock import Mock, patch
+
+from papermill.exceptions import PapermillException
```
MSeal: Let's use relative imports (as is now in master)

harph: ok

PR #83, opened 2018-01-09
Added backslash escaping to prevent python parse errors on notebook execution with errant backslashes

MSeal:
The large number of backslashes are to escape the escaped strings within the test definitions. Tested each locally that print(foo) displays the correct number of `\` characters for these tests.

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -229,6 +229,13 @@ def _find_parameters_index(nb):
         raise PapermillException("Multiple parameters tags found")
     return parameters_indices[0]
 
+def _form_escaped_value(str_val):
+    if isinstance(str_val, string_types):
+        str_val = str_val.encode('unicode_escape')
+        if sys.version_info >= (3, 0):
+            str_val = str_val.decode('utf-8')
+        str_val = str_val.replace('"', r'\"')
+    return '"{}"'.format(str_val)
 
 # Registry for functions that build parameter assignment code.
 _parameter_code_builders = {}
@@ -243,15 +250,13 @@ def register_param_builder(name):
 
     return wrapper
 
-
 @register_param_builder("python")
 def build_python_params(parameters):
     """Writers parameter assignment code for Python kernels."""
     param_content = "# Parameters\n"
     for var, val in parameters.items():
-        if isinstance(val, string_types):
-            val = '"%s"' % val.replace('\\', '\\\\').replace('"', r'\"')
-        param_content += '%s = %s\n' % (var, val)
+        val = _form_escaped_value(val)
+        param_content += '{} = {}\n'.format(var, val)
     return param_content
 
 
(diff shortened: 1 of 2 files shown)


papermill/tests/test_execute.py
```diff
@@ -45,6 +45,23 @@ def test_quoted_params(self):
...
+        execute_notebook(self.notebook_path, self.nb_test_executed_fname, {'foo': r'bar=\"baz\"'})
+        test_nb = read_notebook(self.nb_test_executed_fname)
+        self.assertEqual(test_nb.node.cells[1].get('source'), u'# Parameters\nfoo = "bar=\\\\\\"baz\\\\\\""\n')
+        self.assertEqual(test_nb.parameters, {'foo': 'bar=\\\"baz\\\"'})
+
+    def test_double_backslash_quote_params(self):
+        execute_notebook(self.notebook_path, self.nb_test_executed_fname, {'foo': r'\\"bar\\"'})
+        test_nb = read_notebook(self.nb_test_executed_fname)
+        self.assertEqual(test_nb.node.cells[1].get('source'), u'# Parameters\nfoo = "\\\\\\\\\\"bar\\\\\\\\\\""\n')
+        self.assertEqual(test_nb.parameters, {'foo': '\\\\\"bar\\\\\"'})
```
rgbkrk: These tests might be a little more readable with a raw string (or do you have to double escape there too)?

rgbkrk:
I've been doing some thinking on this and am curious if we should be using Python's built-in `decode`:

```
myString = u'# Parameters\nfoo = "do\\\\ not\\\\ crash"\n'
decoded_string = bytes(myString, "utf-8").decode("unicode_escape") # python3 
# decoded_string = myString.decode('string_escape') # python2
decoded_string
```

screen shot 2018-01-09 at 11 21 54 am

papermill/tests/test_execute.py
MSeal: It's one less level of escaping, but the parameters with `\n` can't use raw strings and match correctly. Though I could split on `\n` and compare lists.

MSeal:
That's not a bad idea; using encodings rather than decodings. It performs the escapse on `\` but not the extra quote escapes we have now. Could combine the two as:
Python 2
py2_encode

Python 3
py3_encode

Think we should go with the encode variant with the extra double quote escape for injection to be a bit safer on conversion?

rgbkrk:
Yeah, the reason I'm thinking we should use the built in encode is to cover any other cases we might be missing / not thinking about while also lowering the maintainability burden of a bunch of `.replace`s.

papermill/execute.py
```diff
@@ -229,6 +229,13 @@ def _find_parameters_index(nb):
         raise PapermillException("Multiple parameters tags found")
     return parameters_indices[0]
 
+def _form_escaped_value(str_val):
+    if isinstance(str_val, string_types):
+        str_val = str_val.encode('unicode_escape')
```
rgbkrk: Does this need to be `string_escape` for Python 2?

MSeal: Nope; see the image above with `unicode_escape` used in python 2. `string_escape` will actually add an extra escape character to the `'` quotes which we don't want either.

rgbkrk: ah, great, thanks!

papermill/execute.py
```diff
@@ -260,13 +265,12 @@ def build_r_params(parameters):
...
-            val = '"%s"' % val.replace('"', '\\"')
-        elif val is True:
+        val = _form_escaped_value(val)
+        if val is True:
             val = 'TRUE'
         elif val is False:
             val = 'FALSE'
-        param_content += '%s = %s\n' % (var, val)
+        param_content += '{} = {}\n'.format(var, val)
     return param_content
```
rgbkrk: Side note, since codecov shows me coverage info inline on GitHub -- this section of code is not covered. May want some basic tests here.

rgbkrk:
Want to go ahead and ship this as a patch release?

MSeal:
Yep will do

PR #85, opened 2018-01-12
[MRG] Add tagging mechanism for default parameters cell

### LINKED ISSUE #84: Clearly tag original parameter cells and inserted parameter cells
Follow on to the work from https://github.com/nteract/papermill/pull/74 to more clearly identify the original cell that was tagged as a parameters cell and the inserted cell with passed parameters.

betatim:
Fixes #84

The cell containing the default parameters of a notebook is now tagged
with "default parameters".

* [x] test basic tagging behaviour
* [x] test existing tags other than "parameters" are kept

WDYT?

### DIFF
--- a/papermill/tests/test_parameterize.py
+++ b/papermill/tests/test_parameterize.py
@@ -1,17 +1,20 @@
+import sys
 import unittest
 
 from ..api import read_notebook
 from ..execute import _parameterize_notebook
 from . import get_notebook_path
 
+PYTHON = 'python2' if sys.version_info[0] == 2 else 'python3'
 
-class TestNotebookHelpers(unittest.TestCase):
+
+class TestNotebookParametrizing(unittest.TestCase):
     def test_preserving_tags(self):
         # test that other tags on the parameter cell are preserved
         test_nb = read_notebook(get_notebook_path("simple_execute.ipynb"))
         test_nb.node.cells[0]['metadata']['tags'].append('some tag')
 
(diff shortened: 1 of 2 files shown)


betatim:
Let me know what you think.

papermill/tests/test_parameterize.py
```diff
@@ -0,0 +1,47 @@
...
+        test_nb = read_notebook(get_notebook_path("simple_execute.ipynb"))
+        test_nb.node.cells[0]['metadata']['tags'].append('some tag')
+
+        _parameterize_notebook(test_nb.node, PYTHON, {'msg': 'Hello'})
+
+        cell_zero = test_nb.node.cells[0]
+        self.assertTrue('some tag' in cell_zero.get('metadata').get('tags'))
+
+        cell_one = test_nb.node.cells[1]
+        self.assertTrue('some tag' in cell_one.get('metadata').get('tags'))
```
MSeal: Hmm do you think having all tags replicated in both cells may cause problems for certain workflows? I.e. workflows that search for a particular flag. I think maybe we should leave tags on the original cell and just move parameters to the new cell, leaving default_parameters in its place? Thoughts?

betatim: I couldn't decide whether as a user I'd expect tags to "propagate" or not :man_shrugging: 

Sounds like you'd expect them not to propagate so going to change it :)

betatim:
Ready.

PR #87, opened 2018-01-13
add cli tests for click

willingc:
- try testing using click's test cli runner

### DIFF
--- a/papermill/tests/notebooks/s3/s3_in/s3-simple_notebook.ipynb
+++ b/papermill/tests/notebooks/s3/s3_in/s3-simple_notebook.ipynb
@@ -1,0 +1,56 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "tags": [
+     "parameters"
+    ]
+   },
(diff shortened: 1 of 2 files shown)


rgbkrk:
Awesome! I'm so glad to see tests getting instrumented for the CLI, thank you @willingc! Is this still WIP? I see the Travis Greenlight so I thought I'd ask.

MSeal:
I coincidentally also [made CLI tests](https://github.com/nteract/papermill/pull/88) today before looking here. Looks like these actually go into calling execute though which is great!

willingc:
@rgbkrk @MSeal These are still a work in process. A fully functioning notebook would be helpful for testing. 

@MSeal More tests are better than not enough ;-)

willingc:
Removing the WIP. Click was giving a bunch of warnings about using the future import so I removed and made the emojis explicit u" ".

rgbkrk:
It looks like a rebase is necessary for this one, otherwise we should go ahead and bring in the enhanced test coverage here.

willingc:
@rgbkrk Rebased :-)

PR #90, opened 2018-01-14
[WIP] Annotate s3 and add basic tests

willingc:


willingc:
Going to try to break this into two PRs - one with tests and one with source. I'm not a big fan of both in one PR.

willingc:
closing since replaced by #100 and #99

PR #88, opened 2018-01-14
Changed yaml parsing to preserve types in notebook. Made multiple yaml inputs valid.

MSeal:
Yaml inputs had to be composed before calling papermill when there didn't seem to be a strict reason to limit inputs. This change makes it so you can mix and match the command inputs to fit your incoming parameter data. In my case I had multiple json objects as well as individual parameters being collected that needed separate assignments in the notebook. Now later yamls will override earlier ones with individual parameters taking highest preference.

Furthermore, passing a json object will give you a dict in your notebook instead of a string. This makes passing complex parameters very nice for consumption in the notebook.

Also added specs for R parameter decoding (and validated they worked even though there aren't R kernel automated specs yet in the repo).

PR #91, opened 2018-01-15
MultiParameter cli

MSeal:
Yaml inputs had to be composed before calling papermill when there didn't seem to be a strict reason to limit inputs. This change makes it so you can mix and match the command inputs to fit your incoming parameter data. In my case I had multiple json objects as well as individual parameters being collected that needed separate assignments in the notebook. Now later yamls will override earlier ones with individual parameters taking highest preference.


papermill/tests/test_cli.py
```diff
@@ -5,49 +5,205 @@
 
 """ Test the command line interface """
 
-import pytest
```
willingc: Let's leave the pytest import.

papermill/tests/test_cli.py
```diff
@@ -5,49 +5,205 @@
...
+from click.testing import CliRunner
+
+class TestCLI(unittest.TestCase):
+    def setUp(self):
+        self.runner = CliRunner()
+        self.default_args = ['input.ipynb', 'output.ipynb']
+        self.sample_yaml_file = os.path.join(os.path.dirname(__file__), 'parameters', 'example.yaml')
+        self.sample_json_file = os.path.join(os.path.dirname(__file__), 'parameters', 'example.json')
+
+    def test_resolve_types(self):
```
willingc: Consider using pytest style for these type tests as was done before.

papermill/tests/test_cli.py
```diff
@@ -5,49 +5,205 @@
...
+        in_out_pairs = [
+            ("True", True),
+            ("False", False),
+            ("None", None),
+            ("12.51", 12.51),
+            ("10", 10),
+            ("hello world", "hello world"),
+            ("😍", "😍"),
+        ]
+        for tin, tout in in_out_pairs:
```
willingc: Within a scope it is better to use simple variable names in Python, `input` `output` and `pairs` as it is easier to read and maintain. Using pytest would give more meaningful test failure output since the runner outputs the specific  test of each pair.

PR #92, opened 2018-01-15
Added python type preservation in parameter passing

MSeal:
Passing a json object will give you a dict in your Python notebook instead of a string. Same for lists, numbers, and booleans. This makes passing complex parameters very nice for consumption in the notebook.

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -230,7 +230,7 @@ def _find_parameters_index(nb):
     return parameters_indices[0]
 
 def _translate_escaped_str(str_val):
-    '''Reusable by most interpreters'''
+    """Reusable by most interpreters"""
     if isinstance(str_val, string_types):
         str_val = str_val.encode('unicode_escape')
         if sys.version_info >= (3, 0):
@@ -239,7 +239,7 @@ def _translate_escaped_str(str_val):
     return '"{}"'.format(str_val)
 
 def _translate_to_str(val):
-    '''Reusable by most interpreters'''
+    """Reusable by most interpreters"""
     return '{}'.format(val)
 
 # Registry for functions that build parameter assignment code.
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -262,12 +262,13 @@ def _translate_type_python(val):
     """Translate each of the standard json/yaml types to appropiate objects in python."""
     if isinstance(val, string_types):
         return _translate_type_str_python(val)
+    # Needs to be before integer checks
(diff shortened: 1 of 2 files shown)


papermill/tests/test_execute.py
```diff
@@ -9,12 +9,46 @@
...
 
 from ..api import read_notebook
-from ..execute import execute_notebook, log_outputs
+from ..execute import execute_notebook, log_outputs, _form_escaped_python_value
 from ..exceptions import PapermillExecutionError
 from . import get_notebook_path, RedirectOutput
 
 python_2 = sys.version_info[0] == 2
 
+class TestPythonEscapers(unittest.TestCase):
```
willingc: I believe you could refactor this to one test case with parameters using pytest.

papermill/execute.py
```diff
@@ -240,6 +240,20 @@ def _form_escaped_value(str_val):
 # Registry for functions that build parameter assignment code.
 _parameter_code_builders = {}
 
+def _form_escaped_python_value(val):
```
willingc: Let's consider renaming to `_translate_type_python`, `_translate_str_python`, `_translate_list_python`.... See https://github.com/nteract/papermill/pull/93#discussion_r161518590 for rationale.

papermill/tests/test_execute.py
MSeal: Hmm true, seems odd to have lots of mixed pytest and unittest components. Are we planning on converting everything over to pytest or just using whichever is cleaner for particular tests? I'll convert this one, just curious on overall direction you see.

betatim: My vote is pytest all the way.

willingc: @MSeal The preferred approach for new tests would be to use pytest as it is far less verbose and maintainable than unittest. The pytest runner will collect tests from either so we don't need to convert from unittest for existing tests. A good project with very clean tests is `requests`.

papermill/execute.py
```diff
@@ -229,17 +229,52 @@ def _find_parameters_index(nb):
         raise PapermillException("Multiple parameters tags found")
     return parameters_indices[0]
 
-def _form_escaped_value(str_val):
+def _translate_escaped_str(str_val):
+    '''Reusable by most interpreters'''
```
willingc: Style tip: use triple double quotes for docstrings `"""`

MSeal: Opps, yep, fixing.

PR #93, opened 2018-01-15
Added r type preservation in parameter passing

MSeal:
Passing a json object will give you a named list in your R notebook instead of a string. Same for lists, numbers, and booleans. This makes passing complex parameters very nice for consumption in the notebook.

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -12,7 +12,7 @@ from jupyter_client.kernelspec import get_kernel_spec
 from nbconvert.preprocessors import ExecutePreprocessor
 from nbconvert.preprocessors.execute import CellExecutionError
 from nbconvert.preprocessors.base import Preprocessor
-from six import string_types
+from six import string_types, integer_types
 from tqdm import tqdm
 
 from .conf import settings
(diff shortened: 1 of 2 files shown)


MSeal:
The R calls corresponding to the R decode specs:

```
#R version 3.3.2 

library("rjson")

t0 = "foo"
t1 = "{\"foo\": \"bar\"}"
t2 <- fromJSON(t1)

print(t0)
print(t1)
print(t2)

t3 = list("foo" = "bar")
t4 = list("foo" = "\"bar\"")

print(t3)
print(t4)

t5 = list("foo" = list("bar"))
t6 = list("foo" = list("bar" = "baz"))
t7 = list("foo" = list("bar" = "\"baz\""))

print(t5)
print(t6)
print(t7)

t8 = list("foo")
t9 = list("foo", "\"bar\"")

print(t8)
print(t9)

t10 = list(list("foo" = "bar"))
t11 = list(list("foo" = "\"bar\""))

print(t10)
print(t11)

t12 = 12345
t13 = -54321

print(t12)
print(t13)

t14 = TRUE
t15 = FALSE

print(t14)
print(t15)
```

with outputs:

```
[1] "foo"
[1] "{\"foo\": \"bar\"}"
$foo
[1] "bar"

$foo
[1] "bar"

$foo
[1] "\"bar\""

$foo
$foo[[1]]
[1] "bar"


$foo
$foo$bar
[1] "baz"


$foo
$foo$bar
[1] "\"baz\""


[[1]]
[1] "foo"

[[1]]
[1] "foo"

[[2]]
[1] "\"bar\""

[[1]]
[[1]]$foo
[1] "bar"


[[1]]
[[1]]$foo
[1] "\"bar\""


[1] 12345
[1] -54321
[1] TRUE
[1] FALSE
```

MSeal:
Merge https://github.com/nteract/papermill/pull/92 first and I'll resolve conflicts here afterwards

papermill/execute.py
```diff
@@ -229,14 +229,31 @@ def _find_parameters_index(nb):
...
-def _form_escaped_value(str_val):
+def _form_escaped_str(str_val):
     if isinstance(str_val, string_types):
         str_val = str_val.encode('unicode_escape')
         if sys.version_info >= (3, 0):
             str_val = str_val.decode('utf-8')
         str_val = str_val.replace('"', r'\"')
     return '"{}"'.format(str_val)
 
+def _form_escaped_r_value(val):
```
willingc: Something about `_form_escaped_...` feels a bit awkward as a name (probably because form can be a noun or a verb, same with escape). Can we simplify to  `_translate_type`, `_translate_str_r`, `_translate_value_r`,  `_translate_dict_r`...? This way if more languages are added the pattern remains consistent `_translate_{type}_{language}`.

MSeal: Sure, might be a little verbose that way but clear on extensibility and expectations.

papermill/execute.py
```diff
@@ -229,17 +229,54 @@ def _find_parameters_index(nb):
         raise PapermillException("Multiple parameters tags found")
     return parameters_indices[0]
 
-def _form_escaped_value(str_val):
+def _translate_escaped_str(str_val):
+    '''Reusable by most interpreters'''
```
willingc: Tip:  triple double quote `"""`

rgbkrk:
Super great, thanks @MSeal!

Commit message wise, you may want to squash down the repeated commit names in the future (using an interactive rebase). Super cool to see all the activity today from everyone!

MSeal:
@rgbkrk Good point. I forgot to rebase instead of merge when the conflicting PRs were getting resolved.

PR #94, opened 2018-01-15
[MRG] Modify tests to run with 'default' python kernels

betatim:
Remove code that runs a different notebook for python 2 and 3. Instead
deduce from the environment which version of python the kernel should be
and name it appropriately.

It bothered me that I couldn't run the tests locally in a python only environment, no legacy python here :) The way travis is setup we have one build per python version. As a result I propose this change so we always run things with "the default version of the environment" (I don't know a better name).


papermill/tests/notebooks/broken1.ipynb
```diff
@@ -38,7 +38,7 @@
     }
    ],
    "source": [
-    "print \"We're good.\""
+    "print(\"We're good.\")"
```
betatim: this is both legal python and python 2.

papermill/tests/test_execute.py
```diff
@@ -14,11 +16,14 @@
 from . import get_notebook_path, RedirectOutput
 
 python_2 = sys.version_info[0] == 2
+PYTHON = 'python2' if python_2 else 'python3'
+execute_notebook = partial(execute_notebook, kernel_name=PYTHON)
```
betatim: I like this solution because I ❤️ `partial` ... I could be persuaded that this is trying to be too clever (for tests which should be as simple as possible). WDYT?

rgbkrk: This is perfect, I'm happy with it as a functional programming proponent.

PR #97, opened 2018-01-15
Add templates, css and doc text

willingc:


rgbkrk:
Really knockin' this all out of the park @willingc!

PR #99, opened 2018-01-15
Add some s3 tests

willingc:


betatim:
LGTM, a PR that only adds tests must be my second favourite type of PR, right after the one that only deletes code :)

willingc:
@MSeal Do you want to push the green button and send coverage over 60%?

PR #100, opened 2018-01-15
Add docstrings for s3

willingc:


papermill/s3.py
```diff
@@ -88,57 +134,91 @@ def wrapper(*args, **kwargs):
                 except Exception as e:
                     logger.debug('retrying: {}'.format(e))
             else:
-                raise
-
+                raise Exception  # TODO verify what to raise
```
MSeal: Should probably save `e` and raise `e` if not None and `Exception("Unknown Failures")` otherwise?

papermill/s3.py
```diff
@@ -406,36 +493,37 @@ def cat(self,
 
             if size != bytes_read:
                 if err:
-                    raise
+                    raise Exception
```
MSeal: Ditto on raising e from earlier. This might warrant more of a refactor to reduce the function complexity to achieve that here though

PR #101, opened 2018-01-15
PEP8 rolling through

betatim:
Nothing to see here, just some PEP8'ing.

### DIFF
--- a/papermill/tests/test_execute.py
+++ b/papermill/tests/test_execute.py
@@ -8,6 +8,8 @@ import unittest
 
 from functools import partial
 
+import six
+
 import nbformat
 
 from ..api import read_notebook
@@ -15,9 +17,11 @@ from ..execute import execute_notebook, log_outputs, _translate_type_python, _tr
 from ..exceptions import PapermillExecutionError
 from . import get_notebook_path, RedirectOutput
 
-python_2 = sys.version_info[0] == 2
-PYTHON = 'python2' if python_2 else 'python3'
-execute_notebook = partial(execute_notebook, kernel_name=PYTHON)
+
+if six.PY2:
+    execute_notebook = partial(execute_notebook, kernel_name='python2')
+else:
+    execute_notebook = partial(execute_notebook, kernel_name='python3')
 
 
 @pytest.mark.parametrize("test_input,expected", [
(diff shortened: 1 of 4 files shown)


papermill/tests/test_execute.py
```diff
@@ -2,22 +2,25 @@
...
 from . import get_notebook_path, RedirectOutput
 
-python_2 = sys.version_info[0] == 2
-PYTHON = 'python2' if python_2 else 'python3'
-execute_notebook = partial(execute_notebook, kernel_name=PYTHON)
+
+if six.PY2:
+    execute_notebook = partial(execute_notebook, kernel_name='python2')
+else:
+    execute_notebook = partial(execute_notebook, kernel_name='python3')
```
MSeal: simpler as `execute_notebook = partial(execute_notebook, kernel_name='python3' if six.PY2 else 'python2')`

papermill/tests/test_parameterize.py
```diff
@@ -1,11 +1,15 @@
...
 import unittest
 
+import six
+
 from ..api import read_notebook
 from ..execute import _parameterize_notebook
 from . import get_notebook_path
 
-PYTHON = 'python2' if sys.version_info[0] == 2 else 'python3'
+if six.PY2:
```
MSeal: nbd but I think the single line version reads better `PYTHON = 'python2' if six.PY2 else 'python3'`

betatim: 👍

papermill/tests/test_execute.py
betatim: Can we split the difference on this one? I wouldn't ever expect a ternary inside an argument like that and it makes the line super long (as in >79chars which is my religion ;) )

MSeal: That 80 char limit is super irritating in modern coding environments (especially with 4 space python :P). If you really don't like the ternary then at least keep it a bit more DRY and make a kernal_name variable which captures the six logic and only repeat `execute_notebook = partial(execute_notebook, kernel_name=kernel_name)` once

MSeal: Sound alright? If you strong disagree I'm still ok with merging as it's better than it was regardless.

betatim: Done.

Next you are going to tell me some of your friends use tabs :p

PR #107, opened 2018-01-23
Let user choose how to deal with execution result.

jarrekk:
Hi, here I add this to let the user choose how to deal with execution result if I use it in an application, I can use variable to deal with the result, no need to read again from the output file. Thanks.

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -136,7 +136,7 @@ Preprocessor.preprocess = preprocess
 
 
 def execute_notebook(notebook,
-                     output=None,
+                     output,
                      parameters=None,
                      kernel_name=None,
                      progress_bar=True,
@@ -151,8 +151,8 @@ def execute_notebook(notebook,
         progress_bar (bool): Flag for whether or not to show the progress bar.
         log_output (bool): Flag for whether or not to write notebook output to stderr.
     """
-    # print("Input Notebook:  %s" % get_pretty_path(notebook))
-    # print("Output Notebook: %s" % get_pretty_path(output))
+    print("Input Notebook:  %s" % get_pretty_path(notebook))
+    print("Output Notebook: %s" % get_pretty_path(output))
     nb = load_notebook_node(notebook)
 
     # Parameterize the Notebook.
@@ -182,10 +182,10 @@ def execute_notebook(notebook,
     nb.metadata.papermill['exception'] = any(
         [cell.metadata.papermill.get('exception') for cell in nb.cells])
 
-    if output:
-        # Write final Notebook to disk.
-        write_ipynb(nb, output)
-        raise_for_execution_errors(nb, output)
+    # Write final Notebook to disk.
+    write_ipynb(nb, output)
+    raise_for_execution_errors(nb, output)
+    # always return notebook object
     return nb
 
 
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -142,7 +142,6 @@ def execute_notebook(notebook,
                      progress_bar=True,
                      log_output=False):
     """Executes a single notebook locally.
-
     Args:
         notebook (str): Path to input notebook.
         output (str): Path to save executed notebook.
@@ -150,6 +149,9 @@ def execute_notebook(notebook,
         kernel_name (str): Name of kernel to execute the notebook against.
         progress_bar (bool): Flag for whether or not to show the progress bar.
         log_output (bool): Flag for whether or not to write notebook output to stderr.
+
+    Returns:
+         nb (NotebookNode): executed notebook object
     """
     print("Input Notebook:  %s" % get_pretty_path(notebook))
     print("Output Notebook: %s" % get_pretty_path(output))
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -136,7 +136,7 @@ Preprocessor.preprocess = preprocess
 
 
 def execute_notebook(notebook,
-                     output,
+                     output=None,
                      parameters=None,
                      kernel_name=None,
                      progress_bar=True,
@@ -151,8 +151,8 @@ def execute_notebook(notebook,
         progress_bar (bool): Flag for whether or not to show the progress bar.
         log_output (bool): Flag for whether or not to write notebook output to stderr.
     """
-    print("Input Notebook:  %s" % get_pretty_path(notebook))
-    print("Output Notebook: %s" % get_pretty_path(output))
+    # print("Input Notebook:  %s" % get_pretty_path(notebook))
+    # print("Output Notebook: %s" % get_pretty_path(output))
     nb = load_notebook_node(notebook)
 
     # Parameterize the Notebook.
@@ -182,9 +182,11 @@ def execute_notebook(notebook,
     nb.metadata.papermill['exception'] = any(
         [cell.metadata.papermill.get('exception') for cell in nb.cells])
 
-    # Write final Notebook to disk.
-    write_ipynb(nb, output)
-    raise_for_execution_errors(nb, output)
+    if output:
+        # Write final Notebook to disk.
+        write_ipynb(nb, output)
+        raise_for_execution_errors(nb, output)
+    return nb
 
 
 def _parameterize_notebook(nb, kernel_name, parameters):
--- a/.gitignore
+++ b/.gitignore
@@ -99,3 +99,6 @@ ENV/
 
 # mypy
 .mypy_cache/
+
+# pyCharm
+.idea


papermill/execute.py
```diff
@@ -151,8 +151,8 @@ def execute_notebook(notebook,
         progress_bar (bool): Flag for whether or not to show the progress bar.
         log_output (bool): Flag for whether or not to write notebook output to stderr.
     """
-    print("Input Notebook:  %s" % get_pretty_path(notebook))
-    print("Output Notebook: %s" % get_pretty_path(output))
+    # print("Input Notebook:  %s" % get_pretty_path(notebook))
+    # print("Output Notebook: %s" % get_pretty_path(output))
```
betatim: I think it is useful to keep printing the input and output paths. (If we decide it isn't useful we should delete the code instead of commenting it out)

papermill/execute.py
```diff
@@ -182,9 +182,11 @@ def execute_notebook(notebook,
     nb.metadata.papermill['exception'] = any(
         [cell.metadata.papermill.get('exception') for cell in nb.cells])
 
-    # Write final Notebook to disk.
-    write_ipynb(nb, output)
-    raise_for_execution_errors(nb, output)
+    if output:
```
betatim: -> `if output is not None`

papermill/execute.py
```diff
@@ -182,9 +182,11 @@ def execute_notebook(notebook,
...
+        # Write final Notebook to disk.
+        write_ipynb(nb, output)
+        raise_for_execution_errors(nb, output)
```
betatim: with this only executed when we write the notebook, what happens if there is an exception and the notebook isn't written to a file?

betatim:
Hey! The idea is to allow people to use `execute_notebook()` from python to run notebooks? Just wanted to double check because I was unsure after reading the title.

papermill/execute.py
jarrekk: Get it, it doesn't matter much :)

papermill/execute.py
jarrekk: Get it 👍

papermill/execute.py
jarrekk: @betatim Thanks, perhaps I have a wrong indent with `raise_for_execution_errors(nb, output)`

jarrekk:
@betatim The main purpose of this PR is to reduce the time of `papermill.execute_notebook` function, in my case, I need to execute a notebook with parameters, then deal output. But default output is in the output file, I have to read this file again, so here I want to get output directly instead of reading the file.

I will test this at local and perhaps also add some unit tests.

Thanks a ton!

papermill/execute.py
```diff
@@ -182,9 +182,11 @@ def execute_notebook(notebook,
     nb.metadata.papermill['exception'] = any(
         [cell.metadata.papermill.get('exception') for cell in nb.cells])
 
-    # Write final Notebook to disk.
-    write_ipynb(nb, output)
     raise_for_execution_errors(nb, output)
+    if output is not None:
+        # Write final Notebook to disk.
+        write_ipynb(nb, output)
+    return nbformat.writes(nb)
```
rgbkrk: This will write the entire notebook to stdout, correct? We definitely don't want that to always happen.

jarrekk: @rgbkrk Thanks, but what if I want to deal with output again? The only way is to read output file for now. Is there a way to drop file I/O step?

jarrekk: @rgbkrk This use case is in Python API, not in the command line. :)

rgbkrk: Oh silly me, I see!

rgbkrk: In that case, it would be much better to return back the notebook object instead of the raw string? That allows you to strip out exactly what you need without having to deserialize the whole thing again. It should make testing the return much easier as well.

```
if output is not None:
  # Write final Notebook to disk.
  write_ipynb(nb, output)

return nb
```

jarrekk: @rgbkrk Yeah, dealing with notebook object is much easier, I made the change. Please review, thanks!

MSeal:
Two comments:
We'd likely want this argument passed through from the cli as well.
If you rerun a notebook in place you will slowly produce additional `default_parameter` cells over time. You may want to change the logic on injecting parameter cells to replace rather than inject if the prior cell has `default_parameter` tag defined.

papermill/execute.py
```diff
@@ -136,7 +136,7 @@ def log_outputs(cell):
 
 
 def execute_notebook(notebook,
-                     output,
+                     output=None,
```
MSeal: Need a spec addition for this behavior :)

papermill/execute.py
```diff
@@ -69,7 +67,7 @@ def preprocess(self, nb, resources):
             cell.outputs = []
 
     # Execute each cell and update the output in real time.
-    with futures.ThreadPoolExecutor(max_workers=1) as executor:
+    with futures.ThreadPoolExecutor(max_workers=4) as executor:
```
betatim: Why increase this? I thought the reason for using a thread pool here is to move the execution of the notebook to somewhere else, not to run things in parallel?

papermill/execute.py
```diff
@@ -79,7 +77,6 @@ def preprocess(self, nb, resources):
 
         for index, cell in execution_iterator:
             cell.metadata["papermill"]["status"] = RUNNING
-            future = executor.submit(write_ipynb, nb, output_path)
```
MSeal: this change looks wrong; you sure you meant to mess with the future pool execution?

papermill/execute.py
```diff
@@ -151,8 +147,6 @@ def execute_notebook(notebook,
         progress_bar (bool): Flag for whether or not to show the progress bar.
         log_output (bool): Flag for whether or not to write notebook output to stderr.
     """
-    print("Input Notebook:  %s" % get_pretty_path(notebook))
```
MSeal: Please leave these in, you can change get_pretty_path to handle None instead

papermill/execute.py
jarrekk: @betatim Yeah, here I just want to try, I will change it back. :)

papermill/execute.py
jarrekk: @MSeal Thanks for your comment, here if I don't use output option in https://github.com/nteract/papermill/blob/master/papermill/execute.py#L138, I will get output file error, if I remove these code, it would be OK, perhaps I need a better way to deal with this.

papermill/execute.py
jarrekk: This is the same about dealing with the output file.

betatim:
I'm still confused about what the end goal is. I originally thought you want to call `execute_notebook` from a different bit of python code, and then do something with the executed notebook. From your comments it sounds like you want to call the papermill executable from the command-line and then capture stdout though. Maybe you could create a short example (in code) of how you want to use it.

papermill/execute.py
betatim: You need to check if `output_path is None` and not always write.

papermill/execute.py
betatim: From your [comment](https://github.com/nteract/papermill/pull/107#issuecomment-360376791) I don't think you need to remove this. This generates some output on stdout but that doesn't get in the way of what you are trying to do.

papermill/execute.py
jarrekk: @betatim Thanks, I added this now. :)

papermill/execute.py
willingc: Hi @jarrekk, I'm a bit confused why we would have output as None for the default? In general, output will need to go somewhere, and it's likely simpler to be explicit about where the output goes. Would you be able to give some additional context to specifically how you see this working? Thanks!

willingc:
Hi @jarrekk,

> #!/usr/bin/env python
nb_object = papermill.execute_notebook(input_file, parameters=parameters) # without output parameter
metadata  = nb_object.metadata
...

Wouldn't you see the same behavior with:

`nb_object = papermill.execute_notebook(input_file, None, parameters=parameters)`

or possibly:

`nb_object = papermill.execute_notebook(input_file, '/dev/null', parameters=parameters)`

I'm a bit concerned that giving a default of None to output would give users who are new to using this unexpected results (i.e. they would expect the output notebook to be saved somewhere and would if using the defaults not have it saved).

jarrekk:
@willingc 
I get your point, perhaps it should use a way to get the `nb_object` without effect on `output_file`. But function `execute_notebook` has no response, so in `nb_object = papermill.execute_notebook(input_file, '/dev/null', parameters=parameters)`, `nb_object` is empty.

jarrekk:
@willingc I updated `execute.py` make it have a response in function `execute_notebook`. :)

willingc:
@jarrekk Thanks for the updates. I would go ahead and add a line in the docstring indicating what is returned. 

@MSeal @betatim Are you cool with this approach?

papermill/execute.py
```diff
@@ -185,6 +185,8 @@ def execute_notebook(notebook,
     # Write final Notebook to disk.
     write_ipynb(nb, output)
     raise_for_execution_errors(nb, output)
+    # always return notebook object
+    return nb
```
rgbkrk: I'm happy with this approach.

MSeal:
With the rebase I'm +1

jarrekk:
@MSeal Thanks!

willingc:
@jarrekk We're in the home stretch now :-)

Please update the docstring for the function since we use the docstring to generate some documentation. Here's a suggestion of two lines to add to the existing docstring. Thanks!

```
"""Executes a single notebook locally.
 
     Args:
         notebook (str): Path to input notebook.
         output (str): Path to save executed notebook.
         parameters (dict): Arbitrary keyword arguments to pass to the notebook parameters.
         kernel_name (str): Name of kernel to execute the notebook against.
         progress_bar (bool): Flag for whether or not to show the progress bar.
         log_output (bool): Flag for whether or not to write notebook output to stderr.

     Returns:
         nb (NotebookNode): executed notebook object
"""
```

jarrekk:
@willingc Great, many thanks!

willingc:
Thanks @jarrekk for being flexible and collaborative. I'm happy that you contributed. Nicely done. :cake:

betatim:
👍 !

PR #110, opened 2018-01-29
string + string (not string + bytes)

michelorengo:
In Python 3, the following:
`
decoded = (undecoded + s).decode(encoding)
`
will result in a TypeError: must be str, not bytes because s is of type bytes and undecoded is of type string.
The suggested PR will fix this issue.

### DIFF
--- a/papermill/s3.py
+++ b/papermill/s3.py
@@ -467,7 +467,7 @@ class S3(object):
 
                         if encoding and not raw:
                             try:
-                                decoded = (undecoded + s).decode(encoding)
+                                decoded = undecoded + s.decode(encoding)
                                 undecoded = ''
                                 yield decoded
                             except UnicodeDecodeError as ude:


MSeal:
@michelorengo Mind if this comes with a test? :) Looks like we were missing some coverage here.

michelorengo:
Ideally, we would want to use a library to mock boto: either moto (https://github.com/spulec/moto) or placebo (https://github.com/garnaat/placebo) is a good choice. There may be others. Idk. Any preferences?
However, you/we may want to ask yourselves/ourselves whether such module (s3.py) should really exist. It probably makes sense to use a library such as s3fs (https://github.com/dask/s3fs). This is what Pandas uses for S3 and given that this project already has a dependency on Pandas it will not add an "exotic" dependency. Not to say that tests should not be written but if we were to spend time on it we may as well refactor the code and use s3fs.

MSeal:
I generally agree with that approach. I'ved used https://github.com/jubos/fake-s3 in the past with a before hook launch -- but it adds ruby as a dependency to tests so I wouldn't recommend it here.

I haven't used s3fs before but your argument sounds solid. There may be a case to be made to add a minimal test here without too much refactor and then go a bigger PR with the swap-over. I'd leave that judgement call to you but I can probably help with s3fs changes/testing later on as well.

michelorengo:
@MSeal, will work on the test but unfortunately, I won't be able to do very soon: my day-to-day job leaves me little time to work on that and going on vacation soon.

MSeal:
I'll go ahead and merge this for now to get the fix in. Let's discuss s3 interface/testing in a separate issue.

PR #112, opened 2018-01-29
ip_display with list arguments

michelorengo:
objects to be displayed with ip_display (IPython.display.display) are in argument list not keywords.
The suggested changes are to remove the keyword argument "data=" and pass the objects as first argument.
(note that the data= does not work on the latest ipython)

PR #114, opened 2018-01-30
Display args fix

MSeal:
objects to be displayed with ip_display (IPython.display.display) are in argument list not keywords.
data= does not work on the latest ipython

Spec fix included from original PR: https://github.com/nteract/papermill/pull/112


MSeal:
Circumventing normal review process and merging to get a fix shipped. Already discussed the fix with @ewmassey

rgbkrk:
Thank you!

PR #120, opened 2018-02-13
Parameterized start_timeout for kernel

MSeal:
For a special kernel we use the startup time is substantially longer than 60 seconds and we want to be able to parameterize that value.


papermill/cli.py
```diff
@@ -29,9 +29,11 @@
               help="Flag for turning on the progress bar.")
 @click.option('--log-output/--no-log-output', default=False,
               help="Flag for writing notebook output to stderr.")
+@click.option('--start_timeout', type=int, default=60,
+              help="Time in seconds to wait for kernel to start.")
```
rgbkrk: 👍

rgbkrk:
:shipit:

Issue #117, opened 2018-02-01
S3 Interface and Testing


MSeal:
From https://github.com/nteract/papermill/pull/110#pullrequestreview-93455074 I wanted to keep the issue open for discussing improving our s3 code and testing.

Michel:
> Ideally, we would want to use a library to mock boto: either moto (https://github.com/spulec/moto) or placebo (https://github.com/garnaat/placebo) is a good choice. There may be others. Idk. Any preferences?
> However, you/we may want to ask yourselves/ourselves whether such module (s3.py) should really exist. It probably makes sense to use a library such as s3fs (https://github.com/dask/s3fs). This is what Pandas uses for S3 and given that this project already has a dependency on Pandas it will not add an "exotic" dependency. Not to say that tests should not be written but if we were to spend time on it we may as well refactor the code and use s3fs.

Matt:
> I generally agree with that approach. I'ved used https://github.com/jubos/fake-s3 in the past with a before hook launch -- but it adds ruby as a dependency to tests so I wouldn't recommend it here.
> 
> I haven't used s3fs before but your argument sounds solid. There may be a case to be made to add a minimal test here without too much refactor and then go a bigger PR with the swap-over. I'd leave that judgement call to you but I can probably help with s3fs changes/testing later on as well.

PR #121, opened 2018-02-20
Add test for s3 back-end (read, write, listdir) using moto lib.

michelorengo:
(sorry for the delay - just got back from vacation).
This is to address issue #117 
The test coverage is admittedly very limited, and it is for 2 reasons:
1. It looks like the s3.py was coded to handle use cases beyond what iorw.py requires. For instance: the ability to read from an archive/compressed file. Yet there are use cases I don't really understand. For example, while "bucket" is a concept defined by the S3 service, the class "Bucket" can be associated with other services (e.g. SQS or EC2). That makes little sense to me, but I'm not claiming I fully understand the original intent.
2. As discussed before, I suggest that the s3.py be scrapped and replaced with a third-party module such as s3fs (which does a good job in providing a python file API on top of S3/boto3)

Therefore because a) I did not understand some of the complexity/ purposes of the s3.py functions and b) would like this module to be replaced, I wrote the tests for the functions iorw.py actually uses.
Hope this is fine.

### DIFF
--- a/papermill/tests/test_s3.py
+++ b/papermill/tests/test_s3.py
@@ -135,3 +135,73 @@ def test_split_error():
     with pytest.raises(ValueError):
         split('https://foo/bar/baz')
 
+
+
+#The following tests are purposely limited to the exposed interface by iorw.py
+import os.path
+import moto
+
+local_dir = os.path.dirname(os.path.abspath(__file__))
+test_bucket_name = 'test-pm-bucket'
+test_string = 'Hello'
+test_file_path = 'notebooks/s3/s3_in/s3-simple_notebook.ipynb'
+
+with open(os.path.join(local_dir, test_file_path)) as f:
+    test_nb_content = f.read()
+
+no_empty_lines = lambda s: "\n".join([l for l in s.split('\n') if len(l) > 0])
+test_clean_nb_content = no_empty_lines(test_nb_content)
+
+read_from_gen = lambda g: "\n".join(g)
+
+
+@pytest.yield_fixture
+def s3_client():
+    mock_s3 = moto.mock_s3()
+    mock_s3.start()
+    import boto3
+    client = boto3.client('s3')
+    client.create_bucket(Bucket=test_bucket_name)
+    client.put_object(Bucket=test_bucket_name,
+                      Key=test_file_path,
+                      Body=test_nb_content)
+    yield S3()
+    try:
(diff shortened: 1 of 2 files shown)


MSeal:
That's perfect. You've stumbled upon some code that was (blindly?) copied over from an internal repository to handle some netflix specific use-cases. You intuition and suggestions on the topic sound right to me.

Looks like the specs need credentials, does it need anonymous creds passed to the client?

michelorengo:
Indeed you need the AWS credentials (from boto3). There are different ways to pass those credentials (see https://boto3.readthedocs.io/en/latest/guide/configuration.html#guide-configuration) but typically those are stored in the AWS credential file. You have a good point though that if you don't have those the tests on s3 will fail. I guess we could bypass the tests if boto3 cannot be properly initialized.

MSeal:
All the aws libraries have anonymous credentials you can set for testing purposes against mock s3's. I can look up the docs for boto3 in particular tomorrow to link.

michelorengo:
Oh! I did not know that. Thanks!

MSeal:
I think this should satisfy the ability: https://stackoverflow.com/questions/34865927/can-i-use-boto3-anonymously
If that has additional issues we can look at making any other credentials anonymous that it complains about (depends on the mock service).

michelorengo:
I see. I misunderstood what you wrote earlier. For some reason, I thought that "test" credentials were available. I understand, now, that you're talking about anonymous access to S3.

michelorengo:
Rainy afternoon so I decided to tackle this. 
Contrary to what I wrote, moto has no problem mocking the authentication, so the issue was not that "Indeed you need the AWS credentials". 
It took me a long, long time (I was too focused on what I added) but I finally saw that the class S3 was instantiated in a different test ("test_s3_defaults"), and because the instances share a dictionary of AWS sessions, the S3() (no args) returned the non-authenticated session.
The fix was then to make sure that S3() was within a moto mock session.
And, of course, this issue is hidden if one runs the test with an AWS credential file. Ah well...

MSeal:
Awesome work. Thanks for getting that fix in. Did @rgbkrk invite you to the Nteract org yet? We like inviting people who make good contributions.

michelorengo:
Thanks. And yes I was invited (and accepted the invitation). When I have a bit of time, I'll work on integrating s3fs package.

PR #123, opened 2018-03-27
switch to html output error message

rgbkrk:
In locked down environments, markdown no longer runs html inline. This switches the error message cell from markdown to HTML. Additionally, the code is hidden so it appears just as HTML output.

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -388,7 +388,15 @@ def raise_for_execution_errors(nb, output_path):
     if error:
         # Write notebook back out with the Error Message at the top of the Notebook.
         error_msg = ERROR_MESSAGE_TEMPLATE % str(error.exec_count)
-        error_msg_cell = nbformat.v4.new_markdown_cell(source=error_msg)
+        error_msg_cell = nbformat.v4.new_code_cell(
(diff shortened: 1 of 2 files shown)


rgbkrk:
cc @mseal for review :bowtie:

papermill/execute.py
```diff
@@ -388,7 +388,14 @@ def raise_for_execution_errors(nb, output_path):
...
         error_msg = ERROR_MESSAGE_TEMPLATE % str(error.exec_count)
-        error_msg_cell = nbformat.v4.new_markdown_cell(source=error_msg)
+        error_msg_cell = nbformat.v4.new_code_cell(
+            source="%%html\n" + ERROR_MESSAGE_TEMPLATE,
+            outputs=[
+                nbformat.v4.new_output(
+                    output_type="display_data",
+                    data={"text/html": ERROR_MESSAGE_TEMPLATE}
+                )]
+        )
```
charsmith: Is this going to hide the input?

rgbkrk: I still need to stick some metadata on to hide the input.

rgbkrk: Updated this to use the metadata recognized by multiple frontends for hidden cells.

rgbkrk:
Thanks!

PR #131, opened 2018-04-12
Made PapermillIO preserve handler ordering

MSeal:
When handlers schemas overlapped in pattern matching the handler picking was non-deterministic. This preserves handler ordering and allows for existing handlers to be removed.

Also changed the class to no longer carry global state about handlers on the class.

### DIFF
--- a/papermill/tests/test_iorw.py
+++ b/papermill/tests/test_iorw.py
@@ -7,7 +7,71 @@ else:
     from mock import Mock, patch
 
 from ..exceptions import PapermillException
-from ..iorw import HttpHandler
+from ..iorw import HttpHandler, PapermillIO
+
+
+class TestPapermillIO(unittest.TestCase):
+    """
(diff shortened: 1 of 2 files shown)


papermill/tests/test_iorw.py
```diff
@@ -7,7 +7,71 @@
...
+        self.papermill_io.register("fake", self.fake1)
+
+    def test_get_handler(self):
+        self.assertEqual(self.papermill_io.get_handler("fake"), self.fake1)
+
+    def test_get_local_handler(self):
+        with self.assertRaises(PapermillException):
+            self.papermill_io.get_handler("dne")
+
+        self.papermill_io.register("local", self.fake2)
```
betatim: the idea is that a `local` handler should be the fallback if nothing else matches?

MSeal: Yes. This was the behavior before, I just preserved it.

Issue #130, opened 2018-04-06
Hide all input for output notebook


rgbkrk:
From a user:

> It would be cool if papermill had an option to hide code cell input when generating an output notebook. This could be very helpful when generating a nightly report and not wanting to see the code that generates the report.

Simple enough. nteract uses `metadata.inputHidden` as a boolean value to indicate if a cell's input is hidden.

~~I suppose a flag like `--hide-inputs` or something? Maybe if we wanted to be more opinionated about the naming we'd call it `--report-mode` or `--mode=report`?~~

Thus far convergence on flag name is `--report-mode`.

mpacer:
This is functionality already built into nbconvert, on the command line
this would be the flag --TemplateExporter.exclude_input=True

betatim:
I like `--report-mode` or `--mode=report` over `--hide-input`.

cldssty:
Can I take a stab at this?

betatim:
Sure thing! Let us know how it is going and feel free to create a PR marked as "work in progress" as soon as you feel like you have something to share (even if it is a long way from being done).

rgbkrk:
Since we'll only have one mode to begin with, we should just call this `--report-mode` for now.

mpacer:
@cldssty since this is already a traitlet in nbconvert you could get away with adding a flag to the papermill jupyter app and setting the value of that nbconvert flag (which I mention above).

rgbkrk:
I _think_ it would be fine if this was a `JupyterApp`. We were happy with using `click` and weren't aware of the benefits of it being a `JupyterApp`.

cldssty:
#135 
Have made an attempt - please tell me what to add/change. Thanks!

OrKoN:
Hi, I got a question about this feature. I run papermill like this: `papermill --report-mode input.ipynb output.ipynb -p ...` I expected that output.ipynb will have input cells hidden if I open it with the jupyter notebook but actually input cells (e.g. code cells) are not hidden. How is this flag supposed  to work?

MSeal:
@OrKoN This feature was relatively new, all of the tools have adopted it except jupyter classic as far I know. I looked at the code involved there and kicked a comment in https://github.com/jupyter/notebook/issues/534 to see if we can get classic to also respect it soon.

PR #135, opened 2018-05-10
Add option for hiding input

cldssty:
Have made an attempt. Please tell me what to change/add. Thanks!

### DIFF
--- a/papermill/cli.py
+++ b/papermill/cli.py
@@ -36,9 +36,11 @@ from .iorw import read_yaml_file
               help="Flag for writing notebook output to stderr.")
 @click.option('--start_timeout', type=int, default=60,
               help="Time in seconds to wait for kernel to start.")
+@click.option('--report-mode/--not-report-mode', default=False,
+              help="Flag for hiding input.")
 def papermill(notebook_path, output_path, parameters, parameters_raw,
               parameters_file, parameters_yaml, parameters_base64, prepare_only,
-              kernel, progress_bar, log_output, start_timeout):
+              kernel, progress_bar, log_output, start_timeout, report_mode):
     """This utility executes a single notebook on a container.
 
     Papermill takes a source notebook, applies parameters to the source
@@ -64,8 +66,8 @@ def papermill(notebook_path, output_path, parameters, parameters_raw,
                      kernel_name=kernel,
                      progress_bar=progress_bar,
                      log_output=log_output,
-                     start_timeout=start_timeout)
-
+                     start_timeout=start_timeout,
+                     report_mode=report_mode)
 
 def _resolve_type(value):
     if value == "True":
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -23,7 +23,8 @@ def execute_notebook(notebook,
                      kernel_name=None,
                      progress_bar=True,
                      log_output=False,
-                     start_timeout=60):
+                     start_timeout=60,
+                     report_mode=False):
     """Executes a single notebook locally.
     Args:
         notebook (str): Path to input notebook.
@@ -34,6 +35,7 @@ def execute_notebook(notebook,
         progress_bar (bool): Flag for whether or not to show the progress bar.
         log_output (bool): Flag for whether or not to write notebook output to stderr.
         start_timeout (int): Duration to wait for kernel start-up.
+        report_mode (bool): Flag for whether or not to hide input.
 
     Returns:
          nb (NotebookNode): Executed notebook object
@@ -41,6 +43,13 @@ def execute_notebook(notebook,
     print("Input Notebook:  %s" % get_pretty_path(notebook), file=sys.stderr)
     print("Output Notebook: %s" % get_pretty_path(output), file=sys.stderr)
     nb = load_notebook_node(notebook)
+    
+    # Hide input if report-mode is set to True.
+    if report_mode: 
+        for cell in nb.cells:
+            if cell.cell_type == 'code':
+                cell.metadata['jupyter'] = cell.get('jupyter', {})
+                cell.metadata['jupyter']['source_hidden'] = True
 
     # Parameterize the Notebook.
     if parameters:
--- a/papermill/iorw.py
+++ b/papermill/iorw.py
@@ -117,7 +117,7 @@ class S3Handler(object):
     @classmethod
     def pretty_path(cls, path):
         return path
-
+            
 
 # Instantiate a PapermillIO instance and register Handlers.
 papermill_io = PapermillIO()
--- a/papermill/tests/test_execute.py
+++ b/papermill/tests/test_execute.py
@@ -257,7 +257,6 @@ class TestLogging(unittest.TestCase):
             stderr = redirect.get_stderr()
             self.assertEqual(stderr, u'Out [3] --------------------------------\n\n')
 
-
 class TestNBConvertCalls(unittest.TestCase):
 
     def setUp(self):
@@ -276,3 +275,20 @@ class TestNBConvertCalls(unittest.TestCase):
         # Ensure the notebook builds valid html without crashing
         (body, resources) = self.html_exporter.from_notebook_node(test_nb.node)
         self.assertTrue(body.startswith("\n<div"))
+
+class TestReportMode(unittest.TestCase):
+
+    def setUp(self):
+        self.test_dir = tempfile.mkdtemp()
+        self.notebook_name = 'report_mode_test.ipynb'
+        self.notebook_path = get_notebook_path(self.notebook_name)
+        self.nb_test_executed_fname = os.path.join(self.test_dir, 'output_{}'.format(self.notebook_name))
+
+    def tearDown(self):
+        shutil.rmtree(self.test_dir)
+
+    def test_report_mode(self):
+        nb = execute_notebook(self.notebook_path, self.nb_test_executed_fname, report_mode = True)
+        for cell in nb.cells: 
+            if cell.cell_type == 'code':  
+                self.assertEqual(cell.metadata.get('jupyter', {}).get('source_hidden'), True)
--- a/papermill/tests/notebooks/report_mode_test.ipynb
+++ b/papermill/tests/notebooks/report_mode_test.ipynb
@@ -1,0 +1,59 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "tags": [
+     "parameters"
+    ]
+   },
+   "outputs": [],
+   "source": [
+    "a = 1\n",
+    "b = 1"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def sum(x, y):\n",
+    "    return x+y"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "print(sum(a, b))"
+   ]
+  }
+ ],
+ "metadata": {
+  "celltoolbar": "Tags",
+  "kernelspec": {
+   "display_name": "Python 3",
+   "language": "python",
+   "name": "python3"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.6.3"
+  }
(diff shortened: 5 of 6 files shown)


cldssty:
#130

papermill/execute.py
```diff
@@ -147,7 +147,8 @@ def execute_notebook(notebook,
                      kernel_name=None,
                      progress_bar=True,
                      log_output=False,
-                     start_timeout=60):
+                     start_timeout=60,
+                     report-mode=None):
```
rgbkrk: This should be an underscore, otherwise this is a subtraction.

papermill/execute.py
```diff
@@ -156,13 +157,18 @@ def execute_notebook(notebook,
...
 
     Returns:
          nb (NotebookNode): executed notebook object
     """
     print("Input Notebook:  %s" % get_pretty_path(notebook), file=sys.stderr)
     print("Output Notebook: %s" % get_pretty_path(output), file=sys.stderr)
     nb = load_notebook_node(notebook)
+    
+    # Hide input if report-mode is set to True.
+    if report-mode:
```
rgbkrk: `report_mode` here too.

cldssty:
ah right! Will fix. Thanks!

cldssty:
hopefully it is better now

cldssty:
hm... what else is wrong?

rgbkrk:
It's only reporting an error because coverage has gone down (no tests on this new command). I'll have to defer to one of the primary pythonistas for what they'd want next. Thanks for submitting this!

cldssty:
Alright! Looking forward to seeing what is done next!

MSeal:
Two things to add now I think:

cldssty:
Right. Will get on with tests in a bit!

cldssty:
hm... what is going on?

cldssty:
will look into the error tmr!

papermill/cli.py
```diff
@@ -36,6 +36,8 @@
               help="Flag for writing notebook output to stderr.")
 @click.option('--start_timeout', type=int, default=60,
               help="Time in seconds to wait for kernel to start.")
+@click.option('--report-mode/--not-report-mode', default=None,
```
MSeal: You'll want this to be `default=False`

papermill/execute.py
```diff
@@ -147,7 +147,8 @@ def execute_notebook(notebook,
                      kernel_name=None,
                      progress_bar=True,
                      log_output=False,
-                     start_timeout=60):
+                     start_timeout=60,
+                     report_mode=None):
```
MSeal: I'd do `report_mode=False` here too as it's a flag

cldssty: ok thanks!

papermill/cli.py
cldssty: right!

cldssty:
sorry about the multiple commits - will squash in a bit!

cldssty:
I have read the errors, and the following is my understanding:

1. test_cli.py 
it says that the arguments in the call aren't consistent with the expected arguments but I see no difference - what am I missing? 

2. test_execute.py
the test notebook 'input.ipynb' is not found in the path because I've only added the notebooks in my branch - presumably that's an error I can't overcome?

Thanks!

MSeal:
Try pasting the delta into a diff tool, should pop up quickly. Also try adding a debug point (or print statements) in the execute call to make sure you're getting expected values. You can isolate to a single test with https://stackoverflow.com/questions/36456920/is-there-a-way-to-specify-which-pytest-tests-to-run-from-a-file to narrow down the issue. I don't see the problem from a peripheral scan of the code.

MSeal:
For test_execute you do not need to have the notebooks pushed anywhere, just make sure they're in the repo branch and added. Check that `self.notebook_path` is getting set to something correct, but I don't see any obvious issue there either so it's probably a silly typo error someplace.

MSeal:
Awesome, two things left. First, let's squash these commits down. Second is we're discussing this week with a few nteract team members what the best metadata / library contract should be to respect this report mode. Specifically what should the metadata field look like and how do we want nteract exporters to respect metadata contracts moving forward. Today I don't believe any tools will respect `nb.metadata.inputHidden = True`, and I'm assuming you want nteract html exporters to respect this field?

rgbkrk:
The original jupyter discussion about hidden input was here: https://github.com/jupyter/notebook/issues/534

nteract frontends all support https://github.com/nteract/nteract/pull/1697/files

I totally forgot that at the last dev meetings we ended up on a nested path:

https://github.com/jupyter/nbformat/pull/94/files

`metadata.jupyter.source_hidden` and `metadata.jupyter.outputs_hidden`

Let's go with that officially merged stance on `nbformat` and I'll make the necessary changes to nteract to _also_ set those. 🙃

cldssty:
Ok! Thanks. I'll squash later today and go through the links!

mpacer:
@cldssty As long as we're namespacing the metadata to provide this functionality — I'm cool with that. With such a change I'm ok with this PR.

Note that that spec that @rgbkrk linked to is a spec on a _per cell_ basis, and so there is no one field to set for the whole notebook. 

More generally though, think we should consider the long-term plan for additional customisation (like report-mode) for a couple of reasons. First, the proposed solution will be respected by the `nteract` front-end and if you want a static html page (which is what I thought you wanted) you'd need to use commuter (which I think supports this?).

As of right now, `source_hidden` is not respected in `nbconvert` as there was some disagreement on whether that hidden state was actually intended to be the mechanism for specifying this in static exports as opposed to only in the interactive view of the notebook. Right now there is no `--no-input` flag in `nbconvert`, but I plan to make a PR for that later today (as I mentioned in the issue associated with this PR… there is already input removal enabled via the `TemplateExporter.exclude_input`). See https://github.com/spatialaudio/nbsphinx/issues/65#issuecomment-322579486 for more context. 

Part of the reason why there was disagreement around the term hide_input is whether that is intended to remove input altogether or make it so that its visibility can be toggled back and forth (see 
https://mpacer.github.io/hiding_tags_nbconvert/hide_cells_based_on_tags.html for an example, for the code see  https://github.com/mpacer/hiding_tags_nbconvert ).

MSeal:
@cldssty Let us know if you get stuck in the final PR updates

MSeal:
@cldssty Do you think you'll have some time to complete this PR or should we make final touches to get it mergable?

cldssty:
I'm really sorry for the delay! I'll be looking at this this weekend and will ask about any problems I'm encountering. Again really sorry about disappearing and thank you so much for your patience. Apologies for any inconvenience caused!!

MSeal:
No problem, and no need to apologize. We wanted to get this nice capability added before the nteract sprints begin in a week so we'll give special attention to this PR and help with getting it finished.

cldssty:
I've skimmed through the links and these are my thoughts: 

If I stick to using a field for the whole notebook(instead of for each cell), then we'd need to 

1. add the field in the nbformat schema? 

2. then we should probably make it so that if this field is set to True, then for each new code cell created, the `metadata.jupyter.source_hidden` field is set to True? 

Alternatively, I could just make this flag toggle input visibility on a cell-by-cell basis. In that case, I would need to change my code but won't need to add a field to nbformat schema? 

In any case, since @mpacer says that `source_hidden` is not currently respected by nbconvert, we'll need to additionally somehow make this flag set `TemplateExporter.exclude_input` in nbconvert to True as well so that input is hidden for html exports? I need to look into the PR by @mpacer. 

I have yet to find out how to do all of this. If there are any misunderstandings or things that cannot be done, please let me know! 

Thanks :D

cldssty:
Currently have something like this in execute.py:
```
# Hide input if report-mode is set to True.
if report_mode: 
    for cell in nb.cells:
        if cell.cell_type == 'code':
            cell.metadata.jupyter.source_hidden = True
```

cldssty:
It's saying that it doesn't have the 'jupyter' key... was I supposed to do this differently?

rgbkrk:
If it doesn't have the jupyter key, go ahead and add it in, then apply `source_hidden`. We formalized on what the spec would be yet haven't started adopting it. If you set it here, we'll be extra motivated to follow through on the nteract frontend at least. 😄 

We can leave it to nbconvert to implement it without bogging this PR down.

cldssty:
ok shall I just squash the commits? Going to look into making this work with html exports in a bit more detail.

cldssty:
Ok, I'm trying this atm. If this isn't something that people like then I can change it: 

```
if not hasattr(cell.metadata, 'jupyter'):
    cell.metadata['jupyter'] = lambda: None
    setattr(cell.metadata.jupyter, 'source_hidden', False)
    setattr(cell.metadata.jupyter, 'outputs_hidden', False)
```
I'm including `outputs_hidden` because it's been mentioned with `source_hidden`.

cldssty:
I'll keep working on this tmr. 

This is what I've done so far:
1. changed field to `metadata.jupyter['source_hidden']` for cells 
2. currently trying to turn this into `metadata.jupyter.source_hidden` but so far my ideas aren't working

What I've tried so far:
1. setting `metadata.jupyter = lambda: None` and then setting attributes `source_hidden` and `outputs_hidden` for that. The travis-ci errors didn't seem to show any direct cause of the failed build.

2. creating a custom object to set `metadata.jupyter` to. First, I'm not sure whether to put the class definition of this object in iorw.py or not, and secondly, I'm still trying to implement a JSON serialiser for it, but idk if this is the right route to go down. 

Please advise! I will look into the exportation part of this and also if it gets too close to the nteract sprint and I still can't find a good solution, I would be happy to hand this over to someone who can wrap this up before the start of the sprint. You could specify a deadline for this if you like. Until then I'm going to keep trying!

Thanks!

papermill/execute.py
```diff
@@ -34,13 +35,20 @@ def execute_notebook(notebook,
...
     """
     print("Input Notebook:  %s" % get_pretty_path(notebook), file=sys.stderr)
     print("Output Notebook: %s" % get_pretty_path(output), file=sys.stderr)
     nb = load_notebook_node(notebook)
+    
+    # Hide input if report-mode is set to True.
+    if report_mode: 
+        for cell in nb.cells:
+            if cell.cell_type == 'code':
+                cell.metadata.jupyter.source_hidden = True
```
MSeal:
```
cell.metadata['jupyter'] = cell.get('jupyter', {})
cell.metadata.['jupyter']['source_hidden'] = True
```

papermill/iorw.py
```diff
@@ -119,6 +119,29 @@ def pretty_path(cls, path):
...
+class Jupyter(object):
+    """Type of the jupyter attribute of metadata of cells.
+
+    """
+    def __init__(self):
+
+        self.source_hidden = False
+        self.outputs_hidden = False
+
+class Dumper(Jupyter):
```
MSeal: You don't need these two classes

papermill/iorw.py
```diff
@@ -167,6 +190,9 @@ def load_notebook_node(notebook_path):
 
         if not hasattr(cell.metadata, 'papermill'):
             cell.metadata['papermill'] = dict()
+        
+        if not hasattr(cell.metadata, 'jupyter'):
```
MSeal: given the snippet I posted you can skip this check/assignment too

papermill/tests/test_execute.py
```diff
@@ -276,3 +275,20 @@ def test_convert_output_to_html(self):
...
+        self.nb_test_executed_fname = os.path.join(self.test_dir, 'output_{}'.format(self.notebook_name))
+
+    def tearDown(self):
+        shutil.rmtree(self.test_dir)
+
+    def test_report_mode(self):
+        nb = execute_notebook(self.notebook_path, self.nb_test_executed_fname, report_mode = True)
+        for cell in nb.cells: 
+            if cell.cell_type == 'code':  
+                self.assertEqual(cell.metadata.jupyter.source_hidden, True)
```
MSeal: `self.assertEqual(cell.metadata.get('jupyter', {}).get('source_hidden'), True)`

MSeal:
Thanks for coming back to wrap up the PR!

I think leaving metadata with primitive objects (dict/str) will uncomplicated the part you're stuck on and posted come comments to help with making that change. Give that change a whirl and I think you'll be good :)

cldssty:
Thanks for the tips!!

rgbkrk:
This is looking pretty good, I think you can take the WIP title off if you think it's ready. :)

cldssty:
So is `source_hidden` going to be respected by html exporters if it gets adopted by nteract frontends and nbconvert, i.e. something that is probably going to be out of the scope of this PR? 

I was going to do a bit more research later today to see if there is anything else I could do regarding html exports but seems like there isn't?

cldssty:
Thanks for all the tips and being helpful + patient! I'll be looking out for further development in related PRs!

Issue #129, opened 2018-04-04
nbconvert not working after running execute_notebook in papermill


Vednus:
I'm using papermill to parameterize some notebooks which I'm then exporting as html reports using nbconvert. But, I've run into issues with nbconvert erroring out if I use it after running `pm.execute_notebook()`. Specifically, this line is the cause since it hijacks nbconvert's `Preprocessor.preprocess` method: https://github.com/nteract/papermill/blob/master/papermill/execute.py#L141

I ended up getting it to work by saving the original nbconvert preprocess method and then reassigning it after I've used papermill. But, I was wondering, is there a better way?

MSeal:
Hmm yes I think we should instead be extending the `ExecutePreprocessor` class instead of monkey patching the parent class method. I'm not sure if there's subtle implications/assumptions in the repo which made that original change, but there's no comments saying so. We can give it a try and see if things are working.

MSeal:
Hit this issue as well when trying to generate HTML of a papermill notebook node. Will try to get to a fix when I get a chance but any other developers who have a spare cycle this one is pretty irritating when hit.

mpacer:
I just brought this up elsewhere but we can probably make a lot of this nicer by using jupyter machinery(since that’s what nbconvert is using under the hood). I’m not going to be able to put any cycles on this for the next couple weeks but if there is interest, afterward I can take a look at this.

MSeal:
We'll snap a new version here in a week or so when some of the other open PRs merge, but fix is in master now.

PR #136, opened 2018-05-10
Fixed preprocessor inheritance to not monkey patch base methods

MSeal:
Addresses https://github.com/nteract/papermill/issues/129

Test added fails without the change removing the monkey patching.

Extra preprocess code that is copied over can be removed once https://github.com/jupyter/nbconvert/pull/811 is merged.


papermill/preprocess.py
```diff
@@ -0,0 +1,180 @@
...
+            Additional resources used in the conversion process.
+        """
+        path = resources.get('metadata', {}).get('path', '')
+        if path == '':
+            path = None
+
+        # clear display_id map
+        self._display_id_map = {}
+
+        # from jupyter_client.manager import start_new_kernel
```
betatim: delete unused import?

papermill/preprocess.py
```diff
@@ -0,0 +1,180 @@
...
+
+        Returns
+        -------
+        nb : NotebookNode
+            The executed notebook.
+        resources : dictionary
+            Additional resources used in the conversion process.
+        """
+        path = resources.get('metadata', {}).get('path', '')
+        if path == '':
```
betatim: for more pythonicness: `if not path`.

papermill/preprocess.py
```diff
@@ -0,0 +1,180 @@
...
+        self.kc.allow_stdin = False
+        self.nb = nb
+
+        try:
+            nb, resources = self.papermill_preprocess(nb, resources)
+        finally:
+            self.kc.stop_channels()
+            self.km.shutdown_kernel(now=self.shutdown_kernel == 'immediate')
+
+        delattr(self, 'nb')
```
betatim: Using `delattr` makes me take note as it is so rare. Should there be a comment explaining why we assign it on line 90 and then get rid of it again here. Seemingly never having done anything with it?

MSeal: That function was moved without any changes. I can clean it up in the move process.

betatim: Thanks! Tidying as we go along :)

papermill/preprocess.py
mpacer: or if you want it to be strict, `if path is ''` since strings are immutable.

betatim: PEP8 would like you to make use of the falsey-ness of empty strings https://www.python.org/dev/peps/pep-0008/#programming-recommendations not that I am a orthodox PEP8 follower or anything *coughcough* ;)

rgbkrk: We should probably pick one path, so I'll line up with the PEP8 recommendation of `if not path`.

MSeal:
I should be able to apply basic cleanups on code I touched this week, been catching up on tasks from being sick.

papermill/preprocess.py
MSeal: I'm just removing the if statement `path = resources.get('metadata', {}).get('path') or None`

Issue #126, opened 2018-03-28
Feature Request: Dryrun


MSeal:
It would be handy if we had a dryrun mode which parameterized a notebook and saved it to the output path without actually executing the cells. This would enable preparing notebooks for execution elsewhere or with alternative kernels in an upstream process. Should be relatively simple to add.

betatim:
nitpick: might not call the command-line option `--dry-run` as I'd expect that to change absolutely nothing on disk, not even write a new notebook. Maybe `--prepare` or some such?

rgbkrk:
`--apply-parameters`?

MSeal:
`--prepare` or `--apply-parameters` both sounds reasonable... perhaps `--parameterize`, `--params-only`, or `--no-execute` as some other possibilities

PR #137, opened 2018-05-11
Added prepare-only flag

MSeal:
Resolves https://github.com/nteract/papermill/issues/126

This flag add the ability to parameterize but not execute a notebook. Useful for showing a user how parameterization will apply without running any code, or for parameterizing notebooks for other programs to consume / execute.


papermill/execute.py
```diff
@@ -222,6 +218,39 @@ def _parameterize_notebook(nb, kernel_name, parameters):
     after = nb.cells[param_cell_index + 1:]
     nb.cells = before + [newcell] + after
 
+    nb.metadata.papermill['parameters'] = parameters
```
betatim: What does this do? I once upon a time knew but have forgotten :-(

Stores the parameters in the metadata for later inspection?

papermill/tests/test_cli.py
```diff
@@ -137,6 +142,20 @@ def test_parameters_base64(self, execute_patch):
...
+            start_timeout=60)
+
+    @patch(cli.__name__ + '.execute_notebook')
+    def test_prepare_only(self, execute_patch):
+        self.runner.invoke(papermill, self.default_args + ['--prepare-only'])
+        execute_patch.assert_called_with(
+            'input.ipynb',
+            'output.ipynb',
+            {},
+            prepare_only=True,
```
betatim: 🎉

papermill/execute.py
MSeal: Yep. I don't know if we use it much, but in theory you could read another notebook and see how it got parameterized easily this way

PR #138, opened 2018-05-13
Added repeat processing of notebooks to papermill

MSeal:
This is an attempt at adding the ability to repeatedly process notebooks without adding an indefinite number of `default parameter` tagged cells. The idea is that we record a parameter cell used in a prior run, then if we find that cell we choose to replace the cell instead of insert after it. You can see the execution patterns in the tests, but essentially these two pattern follow:

First run, parameter cell set:
execute(['params', None, None], prior_cell = None) => ['default params', 'parameters', None]
Second run, output parameter cell from prior run identified:
execute(['default params', 'parameters', None], prior_cell = 1) => ['default params', 'parameters', None]

First run, no parameter cell set:
execute([None, None, None], prior_cell = None) => ['parameters', None, None]
Second run, output parameter cell from prior run identified:
execute(['parameters', None, None], prior_cell = 0) => ['parameters', None, None]

If a user moves the parameter tag such that it no longer lines up with the prior execution. We assume that the user has modified the notebook outside of papermill and we treat the run as normal execution. If we run an older notebook built by papermill, we default the parameter_cell field to the parameter cell provided we find other indicators of a papermill execution (`parameters` in metadata).

Open to suggested changes, this is indented to better support users who point papermill outputs back to inputs of new papermill executions. It's currently a little confusing when there's 5 `default parameters` cells.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -63,6 +63,10 @@ for the parameters passed in at execution time. It acheive this by inserting a
 cell after the tagged cell. If no cell is tagged with ``parameters`` a cell will
 be inserted to the front of the notebook.
 
+Additionally, you can rerun notebooks through papermill and it will reuse the
+parameter cell it injected in the prior run. When a papermill generated notebook
+executes it will instead replace the ``parameters`` tagged cell.
+
 .. image:: docs/img/parameters.png
 
 Executing a Notebook
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -189,7 +189,6 @@ def execute_notebook(notebook,
     # always return notebook object
     return nb
 
-
 def _parameterize_notebook(nb, kernel_name, parameters):
     """Assigned parameters into the appropiate place in the input notebook
     Args:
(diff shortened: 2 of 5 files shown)


papermill/iorw.py
```diff
@@ -154,18 +154,14 @@ def load_notebook_node(notebook_path):
...
-            'environment_variables': dict(),
             'version': __version__
         }
 
     for cell in nb.cells:
         if not hasattr(cell.metadata, 'tags'):
-            cell.metadata['tags'] = [
-            ]  # Create tags attr if one doesn't exist.
+            # Create tags attr if one doesn't exist.
+            cell.metadata['tags'] = []
```
rgbkrk: Much cleaner to read now. 😄

papermill/execute.py
```diff
@@ -262,16 +285,15 @@ def _build_parameter_code(kernel_name, parameters):
...
     parameters_indices = []
     for idx, cell in enumerate(nb.cells):
-        if "parameters" in cell.metadata.tags:
+        if 'parameters' in cell.metadata.tags:
             parameters_indices.append(idx)
     if not parameters_indices:
         return -1
     elif len(parameters_indices) > 1:
-        raise PapermillException("Multiple cells with parameters tag found")
+        raise PapermillException("Multiple cells with parmeters tag found")
```
rgbkrk: *parameters

README.rst
```diff
@@ -63,6 +63,10 @@ for the parameters passed in at execution time. It acheive this by inserting a
 cell after the tagged cell. If no cell is tagged with ``parameters`` a cell will
 be inserted to the front of the notebook.
 
+Additionally, you can rerun notebooks through papermill and it will reuse the
+parameter cell it injected in the prior run. When a papermill generated notebook
+executes it will instead replace the ``parameters`` tagged cell.
```
rgbkrk: I had to re-parse this mentally multiple times to understand.

What I think might have been confusing for me is that the original model for papermill was to replace the parameters cell to begin with. Then if I recall correctly we started inserting another cell so that we wouldn't have to parse the cell to do a perfect replacement (is that right)?

How does this setup ensure consistency?

papermill/execute.py
```diff
@@ -204,22 +203,29 @@ def _parameterize_notebook(nb, kernel_name, parameters):
...
-    before = nb.cells[:param_cell_index + 1]
-    after = nb.cells[param_cell_index + 1:]
-    nb.cells = before + [newcell] + after
+    param_cell = _get_cell(nb, param_cell_index)
+    meta_param_cell_index = nb.metadata.papermill.get('parameter_cell', param_cell_index)
+
+    # Check if we have a prior run's data to override
+    if 'parameters' in nb.metadata.papermill and \
+            param_cell_index == meta_param_cell_index and \
+            param_cell:
```
rgbkrk: This seems like a dangerous comparison since it relies on ordering. Would it help if we assigned unique IDs to the cells in metadata?

MSeal: Possibly, or we can just not check param cell index at all? My original thinking was that if a user copied an executed notebook and changed it you'd want to treat it as a new notebook.

Actually a better way to achieve this would be to md5 the artifact and save that as a metadata marker marker. Then if a user makes changes the hash would differ.

README.rst
MSeal: Correct. The difference now is that we know the parameter cell is a generated element and perfect replacement is not necessary. Before we had defaults in the `parameters` cell generated by users which the replacement call would erase if it didn't have a new value for that name.

Having thought about the changes and the complexity for a couple days. I think maybe the PR should change to be simpler. Maybe this could be the new rule: If a `default parameters` cell exists, replace the `parameters` cell. If it doesn't exist, change `parameters` to `default parameters` and insert a new `parameters` cell. @rgbkrk What do you think about that model? I think it's easy to describe and still makes rerunning notebooks have a nicer output format.

papermill/execute.py
rgbkrk: That seems sensible, though I haven't fully thought through this.

PR #139, opened 2018-05-18
fixed display with multiple outputs per cell

chriswbishop:


chriswbishop:
Not sure I follow the Codecov failures above - can someone enlighten me?

rgbkrk:
> Not sure I follow the Codecov failures above - can someone enlighten me?

Basically, new lines were added that are not tested so the patch itself isn't well covered and the overall coverage has gone down too. If you're able to add tests, it'll pass codecov.

One of the main leads on this project, @mseal, has been sick lately so it may be a little bit before you get a review (unless @betatim or someone else chimes in).

papermill/api.py
```diff
@@ -221,24 +221,54 @@ def display_output(self, name):
         if name not in outputs:
             raise PapermillException(
                 "Output Name '%s' is not available in this notebook.")
-        output = outputs[name]
-        ip_display(output.data, metadata=output.metadata, raw=True)
+
+        # Loop over elements of output
+        for output in outputs[name]:
```
MSeal: Was there a contract change here we missed? Seems like this would explode if `output[name]` is not an iterable element. Do we need to guard against both patterns (likely)?

papermill/api.py
```diff
@@ -221,24 +221,54 @@ def display_output(self, name):
...
+
+    if _cell_has_output_tag(cell):
+
+        names = set([output.metadata.papermill.get('name') for output in cell.outputs if 'papermill' in output.get('metadata', [])])
+
+        # If there are conflicting names, throw a shoe
+        if len(names) != 1:
+            raise NotImplemented
+
+        return list(names)[0]
```
MSeal: `return next(names)`

papermill/api.py
```diff
@@ -221,24 +221,54 @@ def display_output(self, name):
...
+
+def _get_cell_output_tag(cell):
+
+    if _cell_has_output_tag(cell):
+
+        names = set([output.metadata.papermill.get('name') for output in cell.outputs if 'papermill' in output.get('metadata', [])])
+
+        # If there are conflicting names, throw a shoe
+        if len(names) != 1:
+            raise NotImplemented
```
MSeal: I'd raise a `ValueError` and give a description for why it can't process names

papermill/api.py
```diff
@@ -221,24 +221,54 @@ def display_output(self, name):
...
+            return True
+
+    return False
+
+
+def _get_cell_output_tag(cell):
+
+    if _cell_has_output_tag(cell):
+
+        names = set([output.metadata.papermill.get('name') for output in cell.outputs if 'papermill' in output.get('metadata', [])])
```
MSeal: ditto on outputs being iterable guard

papermill/api.py
```diff
@@ -221,24 +221,54 @@ def display_output(self, name):
...
+def _cell_has_output_tag(cell):
+
+    for output in cell.get('outputs', []):
+        if 'papermill' in output.get('metadata', {}) and output.metadata.papermill.get('name'):
+            return True
+
+    return False
+
+
+def _get_cell_output_tag(cell):
```
MSeal: Isn't this fetching a name instead of a tag?

papermill/api.py
```diff
@@ -221,24 +221,54 @@ def display_output(self, name):
...
+        for output in outputs[name]:
+            ip_display(output.data, metadata=output.metadata, raw=True)
 
 
 def _get_papermill_metadata(nb, name, default=None):
     """Returns a dictionary of a notebook's metadata."""
     return nb.metadata.get('papermill', {}).get(name, default)
 
 
+def _cell_has_output_tag(cell):
```
MSeal: Isn't this checking for a name instead of a tag?

MSeal:
Also, thanks for putting together the PR. I'm healthy again so I should be a bit more responsive to PR changes :)

chriswbishop:
Thank you for feedback, apologies I'm so slow to respond. Got slammed with work, will get to it though. All your comments are spot on, will require changes on my end.

mpacer:
note @chriswbishop if you install a code coverage browser extension it will highlight which of your lines have not been covered by tests in your diff view.

For example (not some line numbers have a red background[uncovered] while others have a green background[covered]) :
image

You can find links to the extensions here: https://docs.codecov.io/docs/browser-extension

papermill/api.py
```diff
@@ -221,24 +221,54 @@ def display_output(self, name):
...
-        for output in cell.get('outputs', []):
-            if 'papermill' in output.get('metadata', {}):
-                output_name = output.metadata.papermill.get('name')
-                if output_name:
-                    outputs[output_name] = output
+
+        name = _get_cell_output_tag(cell)
+
+        if name is not None:
+            outputs[name] = cell.get('outputs', [])
```
mpacer: This should check for the types of cells and only do something if it is a code_cell: https://nbformat.readthedocs.io/en/latest/format_description.html#code-cells 

Then you can assume the existence of outputs as outputs are required for code cells and are not used in any other kind of cell.

https://github.com/jupyter/nbformat/blob/3af03b7804357d967aa035d677c5be64c8a86680/nbformat/v4/nbformat.v4.schema.json#L184
```
            "required": ["cell_type", "metadata", "source", "outputs", "execution_count"],
```

papermill/api.py
mpacer: Is having more than one name in cell outputs not ok? Is that specified anywhere? We specifically have designed the notebook to be able to have multiple cell outputs, so this seems like an odd requirement.

papermill/api.py
mpacer: Note, if you return an iterable of booleans you can directly iterate over the enumerated iterable to give you the conditional you need for the whole set (avoiding the redundant metadata checking

```python
names = set(cell.outputs[idx].metadata.papermill.get('name') 
            for idx, has_name in enumerate(_cell_outputs_with_names(cell)) if has_name)
```

also you don't need to wrap it in a list since you can pass the generator directly to the set in its surrounding parens.

papermill/api.py
```diff
@@ -221,24 +221,54 @@ def display_output(self, name):
...
     outputs = {}
+
     for cell in nb_node.cells:
-        for output in cell.get('outputs', []):
-            if 'papermill' in output.get('metadata', {}):
-                output_name = output.metadata.papermill.get('name')
-                if output_name:
-                    outputs[output_name] = output
+
+        name = _get_cell_output_tag(cell)
```
mpacer: This only works if there is a unique name for cell outputs — can we guarantee that this is always the case?

mpacer: Also — this would only return one output even if there were other outputs that were not named — is that desired?

papermill/api.py
mpacer: I'd prefer for the uniqueness criterion to be implemented one level up (in `_get_notebook_outputs`) rather than in this check. Here you're only getting the names themselves, the logic works whether you return one or many outputs. 

Importantly, by approaching it in this way we gain some forward compatibility. Even if we don't support multiple named outputs in a single cell today, if we were to begin to support that in the future, it won't require changing this API in order to enable that if it already returns an iterable. We'll then need to relax the uniqueness criterion in exactly one place (`_get_notebook_outputs`).

MSeal:
Closing this PR as it's been inactive for a year

PR #141, opened 2018-06-06
Extras require

mpacer:
This adds extras_require to the setup.py and updates the dependency to be more targeted at notebook (not the jupyter metapackage that is less clear from a dependency perspective what it is doing).

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,6 @@
 codecov
 coverage
-jupyter
+notebook
 mock
 pytest>=3.3
 pytest-cov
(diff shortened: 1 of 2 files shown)


rgbkrk:
What's the plan for the next release? This all is looking great, seems like a minor release is in order. 😄

MSeal:
Let's do it now. It's been awhile and the remaining PRs are a little stagnant atm

rgbkrk:
Since I logged in to see who all had access to publishing of papermill, I noticed our original tagline and thought maybe we should update it -- https://github.com/nteract/papermill/pull/142

screen shot 2018-06-07 at 5 12 36 pm

PR #144, opened 2018-06-11
Version fix with pip

mpacer:
This modifies our `requirements.txt` to require `ipython>=5.0` without specifying a Python version. It also makes it such that when a user attempts to install papermill with pip <9.0.1, this will ask them to install pip >= 9.0.1, while explaining that this is a requirement of papermill.

This solution allows us to use IPython's solution for determining whether it should be version 5.x vs version 6+. 
**Note**:  IPython 7.0 will be Python 3.4+ compatible (dropping support for 3.3), and our current approach gets more complicated each time a new compatibility change is released. Instead, I suggest we rely on IPython to do the right thing for IPython.

Additionally, this solution allows us to have a more useful error message if someone is using too old a version of pip to install papermill. It still relies on `IPython`'s logic if someone uses `python setup.py develop`, meaning we do not need to implement secondary checks in our `__init__.py`.

This also makes it work as a universal wheel without requiring any of the checks inside `setup.py`.

To get more information on the way that IPython handles this (and why it handles it this way) you can watch this [talk](https://www.youtube.com/watch?v=2DkfPzWWC2Q) & or read these [slides](https://speakerdeck.com/pycon2017/py3-compatibility-in-a-user-friendly-manner). This ends up being a concrete implementation of the recommendations described in the [practicalities section of the python3 statement](https://python3statement.org/practicalities/).

@Carreau @MSeal @rgbkrk

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,15 +5,12 @@ python:
   - 3.5
   - 3.6
 install:
-  - |
-    python2 -m pip install ipykernel --user "ipython<6"
-    python2 -m pip install ipykernel "ipython<6"
-    python2 -m ipykernel install --user
-    python -m pip install ipykernel "ipython<6"
-    python -m ipykernel install --user
-    pip install -e .
-    pip install -r requirements-dev.txt
+  - pip install setuptools pip --upgrade
+  - pip install -e .[test]
+  - python -m ipykernel install --user
 script:
-  - pytest -v --maxfail=2 --cov=papermill papermill/tests
+  # cd so we test the install, not the repo
+  - cd `mktemp -d`
+  - pytest -v --maxfail=2 --cov=papermill --pyargs papermill
 after_success:
   - codecov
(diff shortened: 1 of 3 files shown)


mpacer:
I iterating on some changes to the travis tests to make it accord better with my understand of what we want to test.

Specifically, I think we don't want to hardcode the python3 should use `ipython<6` and we shouldn't be testing our directory but the installed library itself (which is why I switched to using a temporary directory and `--pyargs papermill`). Most of the changes I'm trying to make are to mirror the simplified version of how [nbconvert handles its tests](https://github.com/jupyter/nbconvert/blob/master/.travis.yml).

Note: I'm not adding the `check-manifest` check because that actually causes it to fail since our MANIFEST.in is not 100% accurate — but that's a separate issue so I'll leave that for another PR.

mpacer:
Note: with the testing changes we now get IPython 6.4 on python3 and IPython 5.7 on python2.

PR #145, opened 2018-06-12
move log_outputs to preprocess

cHYzZQo:
This allows it to be used from both preprocess and execute.
    
Fixes errors like

```
    NameError: global name 'log_outputs' is not defined
```


MSeal:
Thanks for getting this fix in -- missed this import and it looks like we had a hole in test coverage.

Issue #148, opened 2018-06-13
Scala parameter mapping needs to indicate long status for larger integers


MSeal:
```
// Parameters
val RUN_TS = 1528866180240
==========================================
Name: Compile Error
Message: <console>:5: error: integer number too large
val RUN_TS = 1528866180240
             ^
<console>:6: error: ';' expected but 'val' found.
val RUN_DATE_MINUS_1 = 20180612
^

StackTrace: 
```

Should instead populate as `val RUN_TS = 1528866180240L`

PR #149, opened 2018-06-13
Added long translations for integer values above max and min scala Int

MSeal:
Solves https://github.com/nteract/papermill/issues/148

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -233,15 +233,18 @@ def _translate_type_r(val):
     return _translate_escaped_str(val)
 
 _translate_type_str_scala = _translate_escaped_str
-_translate_type_int_scala = _translate_to_str
 _translate_type_float_scala = _translate_to_str
+
+def _translate_type_int_scala(val):
+    strval = _translate_to_str(val)
+    return strval + "L" if val > 2147483647 or val < -2147483648 else strval
 
 def _translate_type_bool_scala(val):
     return 'true' if val else 'false'
 
 def _translate_type_dict_scala(val):
     """Translate dicts to scala maps"""
-    escaped = ', '.join(["{} -> {}".format(_translate_type_str_scala(k), _translate_type_scala(v)) 
+    escaped = ', '.join(["{} -> {}".format(_translate_type_str_scala(k), _translate_type_scala(v))
                          for k, v in val.items()])
     return 'Map({})'.format(escaped)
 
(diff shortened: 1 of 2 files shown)


papermill/execute.py
```diff
@@ -233,15 +233,18 @@ def _translate_type_r(val):
     return _translate_escaped_str(val)
 
 _translate_type_str_scala = _translate_escaped_str
-_translate_type_int_scala = _translate_to_str
 _translate_type_float_scala = _translate_to_str
 
+def _translate_type_int_scala(val):
+    strval = _translate_to_str(val)
+    return strval + "L" if val > 2147483647 or val < -2147483648 else strval
```
mpacer: You could do `abs(val+.5) > 21474783647.5`, either that or I'd recommend putting parents around the condition for legibility.

papermill/execute.py
```diff
@@ -233,15 +233,18 @@ def _translate_type_r(val):
     return _translate_escaped_str(val)
 
 _translate_type_str_scala = _translate_escaped_str
-_translate_type_int_scala = _translate_to_str
 _translate_type_float_scala = _translate_to_str
```
mpacer: Should we have a similar rule for floats vs. doubles?

mpacer: I guess that would be more of an optimization than fixing a limitation. Since doubles are the default & can cover the range that floats support.

papermill/execute.py
mpacer: Should we have [BigInt](https://www.scala-lang.org/api/current/scala/math/BigInt.html) as a fallback for numbers longer than Long supports?

MSeal: Since these come from json spec inputs, https://tools.ietf.org/html/rfc7159#section-6 says Long will be sufficient.

MSeal: I'll wrap the conditional in parens. I find the two conditionals much easier to quickly grok than the `abs(val+.5) > 21474783647.5` where I'd be worried about floating rounding errors.

papermill/execute.py
MSeal: Yep. I think it's fine as is.

papermill/execute.py
mpacer: I don't think there should be rounding errors when comparing two floats, but I agree that it's easier to quickly grok.

PR #151, opened 2018-06-13
verify that the wheel can be installed

rgbkrk:
One quick integration test. It would be better to set up a virtualenv at the very end here (doing the right thing for Python2 vs Python3), this does at least verify that the `setup.py` runs in place within a distributed wheel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ script:
   - pytest -v --maxfail=2 --cov=papermill --pyargs papermill
   - cd $TRAVIS_BUILD_DIR
   - python setup.py sdist bdist_wheel
-  - python -m pip install dist/papermill*.whl
-  - python -m pip install dist/papermill*.tar.gz
+  - python -m pip install --force-reinstall dist/papermill*.whl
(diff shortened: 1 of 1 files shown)


rgbkrk:
Feel free to reject these changes by the way in favor of a better approach that does an isolated environment. I will have no hard feelings over a closed PR, my primary intent is some integration testing 😄

MSeal:
We want to add a check that it can install the tar.gz too?

rgbkrk:
> We want to add a check that it can install the tar.gz too?

Likely a good idea to make sure it can install from source. Not sure how to do that.

MSeal:
`python -m pip install dist/papermill*.tar.gz`

mpacer:
The wheel I don't think was the problem for 0.13.2, but rather the sdist was.

mpacer:
However… i want to note that when i tried to make this fail locally… it did not fail, so I don't know if this is really doing enough.

mpacer:
(sorry about all the small commits — normally i'd force push but I really hate doing that on an upstream branch)

rgbkrk:
What did you do to try to make it fail?

PR #153, opened 2018-06-29
Made http handler raise_for_status

MSeal:
Discovered that setting an output path to an invalid location did not raise an error in papermill execution.

### DIFF
--- a/papermill/tests/test_iorw.py
+++ b/papermill/tests/test_iorw.py
@@ -1,6 +1,7 @@
 import json
 import unittest
 import six
+from requests.exceptions import ConnectionError
 if six.PY3:
     from unittest.mock import Mock, patch
 else:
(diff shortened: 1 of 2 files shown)


papermill/tests/test_iorw.py
```diff
@@ -106,12 +107,22 @@ def test_read(self):
...
             mock_put.assert_called_once_with(path, json=json.loads(buf))
+
+    def test_write_failure(self):
+        """
+        Tests that the `write` function raises on failure to put the buffer.
+        """
+        path = 'http://localhost:9999'
+        buf = '{"papermill": true}'
+
+        with self.assertRaises(ConnectionError) as e:
```
mpacer: could you use `pytest`'s `pytest.raises(ConnectionError)` context manager and match on the message including the failed path in question?

MSeal: I could, mostly I wanted to just prove that request exceptions are raised and not ignored -- I care less what their contents are

MSeal: If you want me to make that change I'll refine the test to be more specific

rgbkrk: Think it's ok to just make an issue for making it more specific about handling an exact message match?

PR #155, opened 2018-07-14
resolves #154 added CONTRIBUTING.md

### LINKED ISSUE #154: Need CONTRIBUTING.md
A guide for new contributors would help with sprints & all kinds of new contributors.

wsavran:
i updated setup.py to allow installing the dev environment with the `pip install papermill[dev]` in addition to `pip install papermill[test]` syntax. it seemed more clear when writing up the instructions, but is definitely not necessary.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ to join our community.
 
 We are an open and friendly community. Everybody is welcome.
 
-We encourage friendly discussions and respect for all. No exceptions.
+We encourage friendly discussions and respect for all. There are no exceptions.
 
 All contributions are equally important. Documentation, answering questions, and
 fixing bugs are equally as valuable as adding new features.
@@ -39,15 +39,22 @@ source dev/bin/activate
 pip install -e .
 ```
 
+_Note: When you are finished you can use `source deactivate` to go back to your base environment._
+
 ### Running Tests Locally
-We need to install the development package before we can run the tests.
+We need to install the development package before we can run the tests. If anything is confusing below,
+always resort to the relevant documentation.
 ```buildoutcfg
 pip install papermill[dev]
 py.test --pyargs papermill
 ```
-If the installation worked properly the tests will pass.
+
+Now there should be a working and editable installation of Papermill to start making your own contributions.
 ## So You're Ready to Pull Request
-The general workflow for this will be: Run Local Test 
+The general workflow for this will be:
+1. Run local tests 
+2. Pushed changes to your forked repository
+3. Open pull request to main repository
 
 ### Run Tests Locally
 ```buildoutcfg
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+## Our Community Values
+
+We are an open and friendly community. Everybody is welcome.
+
+We encourage friendly discussions and respect for all. There are no exceptions.
+
+All contributions are equally important. Documentation, answering questions, and
+fixing bugs are equally as valuable as adding new features.
+
+Check out <https://www.python.org/psf/codeofconduct/> for the Python code of conduct.
```
mpacer: nteract has its own code of conduct, so we should probably link to that: https://github.com/nteract/nteract/blob/master/CODE_OF_CONDUCT.md

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+. activate env
+``` 
+or [virtualenv](https://packaging.python.org/guides/installing-using-pip-and-virtualenv/) if you prefer.
+```
+python3 -m virtualenv dev
+source dev/bin/activate 
+```
+
+```
+pip install -e .
```
mpacer: If we're going to encourage a dev install, then we should probably have people install their tests here rather than below. E.g.,

```
pip install -e .[dev]
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+_Note: When you are finished you can use `source deactivate` to go back to your base environment._
+
+### Running Tests Locally
+We need to install the development package before we can run the tests. If anything is confusing below,
+always resort to the relevant documentation.
+```buildoutcfg
```
mpacer: Does this style affect what is rendered? If it doesn't, is there a good reason to continue to include it?

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+pip install papermill[dev]
```
mpacer: if you give the initial instructions to install editably with the dev `extra_requires` then you shouldn't need this (also, it will avoid any issues where the released version differs in its dev requirements from the local version).

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+Now there should be a working and editable installation of Papermill to start making your own contributions.
+## So You're Ready to Pull Request
+The general workflow for this will be:
+1. Run local tests 
+2. Pushed changes to your forked repository
+3. Open pull request to main repository
+
+### Run Tests Locally
+```
+py.test --pyargs papermill
```
mpacer: While you can also run pytest from anywhere within the directory, this actually tests the module and is following pytest's [best practices](https://docs.pytest.org/en/latest/goodpractices.html#tests-as-part-of-application-code) which I hadn't realised was a best practice (just a useful way of being able to run this anywhere). 

I would include that link with a brief explanation of what `--pyargs` is doing when you run the tests.

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+Run check manifest to ensure all files are accounted for in the repository.
```
mpacer: I would add a brief description of what `MANIFEST.in` is doing in python packaging.

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+1. Commit changes to local repository:
+    ```
+    git checkout -b my-feature
+    git add <updated_files>
+    git commit
+    ```
+2. Push changes to your remote repository:
+    ```
+    git push -u origin my-feature 
+    ```
```
mpacer: Can you add a new line before each new section?

PR #163, opened 2018-08-01
Changed parameter management to add an injected-parameters cell. 

MSeal:
Made execution reuse the injected-parameters cell for repeated calls to the same output notebook

Remake of https://github.com/nteract/papermill/pull/138/files

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -43,9 +43,9 @@ def execute_notebook(notebook,
     print("Input Notebook:  %s" % get_pretty_path(notebook), file=sys.stderr)
     print("Output Notebook: %s" % get_pretty_path(output), file=sys.stderr)
     nb = load_notebook_node(notebook)
-    
+
     # Hide input if report-mode is set to True.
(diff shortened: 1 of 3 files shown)


papermill/execute.py
```diff
@@ -90,19 +90,25 @@ def _parameterize_notebook(nb, kernel_name, parameters):
     kernel_name = kernel_name or nb.metadata.kernelspec.name
     param_content = _build_parameter_code(kernel_name, parameters)
 
-    param_cell_index = _find_parameters_index(nb)
-    if param_cell_index >= 0:
-        old_parameters = nb.cells[param_cell_index]
-        old_parameters.metadata['tags'].remove('parameters')
-        old_parameters.metadata['tags'].append('default parameters')
```
rgbkrk: When we see the old tag should we take it into account or warn about it?

MSeal: I think it's actually ok, because it will behave as it did before.

PR #166, opened 2018-08-02
Stream outputs when logging

mpacer:
This changes the order in which we are processing messages when running individual cells. 

Importantly, we are no longer waiting for `execute_reply` which means we can no longer expect all iopub messages to be present when processing them.

While this is a useful simplifying assumption, it means we cannot stream the outputs as they come in (behaviour expected from other front-end notebook contexts).

To address this change in the expectations around the iopub messages, I increased the default iopub_timeout value. We may want to strengthen this and disable iopub timeout errors by default.

Because we don't know what the execution count is until we get the execute reply, we cannot canonically state what output number these cells will have, so we optimistically try to log the streamed messages.

The nbconvert execute preprocessor would benefit from further refactoring so that we did not have to recreate so much of the code… but as of right now that's not the case.

### DIFF
--- a/papermill/preprocess.py
+++ b/papermill/preprocess.py
@@ -7,6 +7,7 @@ from concurrent import futures
 from nbconvert.preprocessors import ExecutePreprocessor
 from nbconvert.preprocessors.execute import CellExecutionError
 from nbformat.v4 import output_from_msg
+from traitlets import default
 
 from .iorw import write_ipynb
 
@@ -72,6 +73,14 @@ class PapermillExecutePreprocessor(ExecutePreprocessor):
 
     # TODO: Delete this wrapper when nbconvert allows for setting preprocessor
     # hood in a more convienent manner
+
+    @default("iopub_timeout")
+    def _default_iopub_timeout(self):
+        """The time to wait (in seconds) for IOPub output.
+        """
+        return 15
+        
+    
     def preprocess(self, nb, resources):
         """
         Copied with one edit of super -> papermill_preprocessor from nbconvert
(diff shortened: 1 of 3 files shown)


papermill/preprocess.py
```diff
@@ -202,3 +194,81 @@ def papermill_preprocess(self, nb, resources):
...
+            if 'execution_count' in content:
+                cell['execution_count'] = content['execution_count']
+
+            if msg_type == 'status':
+                if content['execution_state'] == 'idle':
+                    break
+                else:
+                    continue
+            elif msg_type == 'execute_input':
+                sys.stdout.write('Executing Cell {:-<40}\n'.format(content.get("execution_count", "*")))
```
MSeal: Need an `if self.log_output: ` here so we only log this when we have set the --log-outputs option.

MSeal: This should also use stderr instead of stdout as it's a message instead of data (for good unix-like command support)

papermill/preprocess.py
```diff
@@ -202,3 +194,81 @@ def papermill_preprocess(self, nb, resources):
...
+                output_idx_list = cell_map.setdefault(cell_index, [])
+                output_idx_list.append(len(outs))
+
+            if self.log_output:
+                log_output(out)
+            outs.append(out)
+
+        exec_reply = self._wait_for_reply(msg_id, cell)
+        if self.log_output: 
+            sys.stdout.write('Ending Cell {:-<43}\n'.format(
```
MSeal: This should use stderr instead of stdout as it's a message instead of data (for good unix-like command support)

Issue #168, opened 2018-08-12
Create a guide for adding new languages to papermill


MSeal:
We should refactor the language extensions to be a plugin system where users can easily write new language or kernel support for mapping parameters. The steps are easy enough but a well written page to guide contributions would be it much more open to new contributors who want support.

PR #172, opened 2018-08-22
Translators

MSeal:
Moved translator functions into a registry pattern (functionality the same)
Moved translator tests + added more tests
Created docs on how to write develop aspects of the app (coupled with changes as a promotion of translators)
- Translators
- IO
- CLI


MSeal:
Second commit is tied to https://github.com/nteract/papermill/issues/168

MSeal:
Merging so it's available before the jupytercon sprint

PR #179, opened 2018-08-26
Added Accept: application/json header to iorw.HttpHandler.read

### LINKED ISSUE #174: Improve reading notebooks from url that need accept json header
Our notebooks are being stored in an [nbgallery](https://github.com/nbgallery/nbgallery) instance.  A generic `requests.get(notebook_url)` returns html.  Papermill's `execute.execute_notebook(notebook_url)` / `iowr.HttpHandler.read` do not offer a way to set a `{'Accept' : 'application/json'}` header in the request.

When we make a request on our own using accept json headers, we can write that notebook to file (`tempfile` or otherwise) and use Papermill as is.  It would be convenient to also be able to pass in a file-like object (`io.StringIO(resp.text)`), json (`resp.text`), and/or dictionary (`resp.json()`) of the downloaded notebook file rather than needing to write to file.

Thanks.

kafonek:
Fixes #174

### DIFF
--- a/papermill/iorw.py
+++ b/papermill/iorw.py
@@ -59,7 +59,7 @@ class PapermillIO(object):
 class HttpHandler(object):
     @classmethod
     def read(cls, path):
-        return requests.get(path).text
+        return requests.get(path, headers={'Accept' : 'application/json'}).text
 
     @classmethod
     def listdir(cls, path):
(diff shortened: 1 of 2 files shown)


MSeal:
I'll merge once I get master build test mocks fixed

PR #184, opened 2018-08-27
add support for azure data lake sources and sinks

VincentBLortie:
Azure data lake sources and sinks are made available using the adl:// prefix.
The path must refer to an subdomain of azuredatalakestore.net

Credentials for the azure data lake will be taken from env variables
(see https://github.com/Azure/azure-data-lake-store-python)

If no credentials are provided through environment variables, a code will
be given to the user for them to login through microsoft.com/devicelogin


VincentBLortie:
I believe an issue with the azure-datastore-lake distribution is what causes the Python 2.7 build to fail. I've opened an issue on their side: https://github.com/Azure/azure-data-lake-store-python/issues/236

MSeal:
Awesome, thanks for adding this. If we really need to we can also make this particular register only available to python 3 since python 2 will be less well supported by surrounding tools moving forward.

VincentBLortie:
Thanks to @akharik's fix in azure-datalake-store, the python 2.7 build now passes 👍

VincentBLortie:
Happy to contribute :) The TechBlog posts are very appreciated!

PR #187, opened 2018-08-28
Add Python 3.7 to travis tests

willingc:

PR #188, opened 2018-08-29
Travis CI: Add Python 3.7 to the testing

cclauss:
#187 done in alignment with travis-ci/travis-ci#9069

Plus whitespace changes in other files to align with #183

On Python 3.7: ModuleNotFoundError: No module named 'google_compute_engine'
* GCE might not yet support Python 3.7...
* https://github.com/GoogleCloudPlatform/compute-image-packages/blob/master/setup.py#L66

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ python:
   - 3.5
   - 3.6
 matrix:
+  allow_failures:  # ModuleNotFoundError: No module named 'google_compute_engine'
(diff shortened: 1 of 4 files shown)


MSeal:
Looks like GCP is had only added 3.7 to a few packages in their latest releases. Even though we don't really use this feature of boto it will block using 3.7 until that gets added over there. I'll add a request to add 3.7 support on their issues.

PR #198, opened 2018-09-02
add unit tests and lint

VincentBLortie:


papermill/s3.py
```diff
@@ -77,7 +77,7 @@ def __str__(self):
         return 's3://{}/{}'.format(self.bucket.name, self.name)
 
     def __repr__(self):
-        return str(self)
+        return self.__str__()
```
VincentBLortie: In s3.py, str() is imported from past.builtins. Because of this, attempting a repr() on either a Key or Prefix object resulted in a TypeError:
```
>>> from papermill.s3 import Prefix
>>> prefix = Prefix("foo", "bar")
>>> prefix
  File "/Users/vincent/workspace/papermill/papermill/s3.py", line 80, in __repr__
    return str(self)
TypeError: cannot convert 'Prefix' object to bytes
```

I am not that familiar with python 2/3 compatibility. This change is what appeared to make the most sense to me.

MSeal: That works, and is forward/backwards compatible

MSeal:
Thanks again for the contributions. Really appreciate getting eyes on test/compatibility improvements like this.

PR #202, opened 2018-09-10
requirements: bump version for nbconvert

stillmatic:
papermill depends on function "_wait_for_reply" which is only available in 5.3 and later, per https://github.com/jupyter/nbconvert/commit/9bd05ffbaefd6b45452cca51fd7ca8964e033c47

not sure if other functions require a newer version.


MSeal:
Thanks for the requirements improvement!

Issue #201, opened 2018-09-08
Nicer error when output directory doesn't exist


betatim:
When the output directory (called `out` in this case) does not exist you end up with the long output below:
```
$ papermill run-me.ipynb out/run-me.ipynb
Input Notebook:  run-me.ipynb
Output Notebook: out/run-me.ipynb
  0%|                                                                                                 | 0/3 [00:00<?, ?it/s]Traceback (most recent call last):
  File "/srv/conda/bin/papermill", line 11, in <module>
    sys.exit(papermill())
  File "/srv/conda/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/srv/conda/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/srv/conda/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/srv/conda/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/srv/conda/lib/python3.6/site-packages/papermill/cli.py", line 97, in papermill
    report_mode=report_mode,
  File "/srv/conda/lib/python3.6/site-packages/papermill/execute.py", line 65, in execute_notebook
    _execute_parameterized_notebook(nb, kernel_name, progress_bar, log_output, start_timeout)
  File "/srv/conda/lib/python3.6/site-packages/papermill/execute.py", line 131, in _execute_parameterized_notebook
    processor.preprocess(nb, {})
  File "/srv/conda/lib/python3.6/site-packages/papermill/preprocess.py", line 99, in preprocess
    nb, resources = self.papermill_preprocess(nb, resources)
  File "/srv/conda/lib/python3.6/site-packages/papermill/preprocess.py", line 202, in papermill_preprocess
    future.result()
  File "/srv/conda/lib/python3.6/concurrent/futures/_base.py", line 425, in result
    return self.__get_result()
  File "/srv/conda/lib/python3.6/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
  File "/srv/conda/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/srv/conda/lib/python3.6/site-packages/papermill/iorw.py", line 170, in write_ipynb
    papermill_io.write(nbformat.writes(nb), path)
[...]

PR #203, opened 2018-09-12
throw nice error if directory doesn't exist

harsham4026:
@MSeal this addresses the issue of Nicer error when output directory doesn't exist #201
Please review the PR and let me know for any changes. At present I added the exception throwing for local file system. Please let me know if I need to do the same for others as well

### DIFF
--- a/papermill/exceptions.py
+++ b/papermill/exceptions.py
@@ -13,6 +13,10 @@ class FileExistsError(AwsError):
     """Raised when a File already exists on S3."""
 
 
+class FileNotFoundError(AwsError):
+    """Raised when a folder doesn't exist"""
+
+
 class PapermillException(Exception):
     """Raised when an exception is encountered when operating on a notebook."""
 
--- a/papermill/exceptions.py
+++ b/papermill/exceptions.py
@@ -32,3 +32,6 @@ class PapermillExecutionError(PapermillException):
         message += "\n"
 
         super(PapermillExecutionError, self).__init__(message)
+
+class NoSuchDirectory(Exception):
+    """Raised when a folder doesn't exist"""
--- a/papermill/exceptions.py
+++ b/papermill/exceptions.py
@@ -32,7 +32,3 @@ class PapermillExecutionError(PapermillException):
         message += "\n"
 
         super(PapermillExecutionError, self).__init__(message)
-
-
-class NoSuchDirectory(Exception):
-    """Raised when a folder doesn't exist"""
--- a/papermill/exceptions.py
+++ b/papermill/exceptions.py
@@ -13,10 +13,6 @@ class FileExistsError(AwsError):
     """Raised when a File already exists on S3."""
 
 
-class FileNotFoundError(AwsError):
-    """Raised when a folder doesn't exist"""
-
-
 class PapermillException(Exception):
     """Raised when an exception is encountered when operating on a notebook."""
 
--- a/papermill/exceptions.py
+++ b/papermill/exceptions.py
@@ -33,5 +33,6 @@ class PapermillExecutionError(PapermillException):
 
         super(PapermillExecutionError, self).__init__(message)
 
+
 class NoSuchDirectory(Exception):
     """Raised when a folder doesn't exist"""
--- a/papermill/iorw.py
+++ b/papermill/iorw.py
@@ -88,12 +88,12 @@ class LocalHandler(object):
 
     @classmethod
     def write(cls, buf, path):
-        path = (path.split("/"))[:-1]
-        path = "/".join(path)
-        if not path.endswith("/"):
-            path += "/"
-        if not os.path.exists(path):
-            raise FileNotFoundError('output folder {} doesn\'t exist!'.format(path))
+        dir = (path.split("/"))[:-1]
+        dir = "/".join(dir)
+        if not dir.endswith("/"):
+            dir += "/"
+        if not os.path.exists(dir):
+            raise FileNotFoundError('output folder {} doesn\'t exist!'.format(dir))
         with io.open(path, 'w', encoding="utf-8") as f:
             f.write(buf)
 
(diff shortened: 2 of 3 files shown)


harsham4026:
@MSeal please review

papermill/iorw.py
```diff
@@ -88,6 +88,12 @@ def listdir(cls, path):
 
     @classmethod
     def write(cls, buf, path):
+        dir = (path.split("/"))[:-1]
```
betatim: I am not quite sure what the next few lines do but I think they try and find the directory name in the path. I'd recommend using https://docs.python.org/3/library/os.path.html#os.path.dirname or other tools from that module to do this.

By using an existing tool from the standard library you make it easier for others to read your code and we are less likely to have a bug in the code that manipulates the path.

papermill/exceptions.py
```diff
@@ -13,6 +13,10 @@ class FileExistsError(AwsError):
     """Raised when a File already exists on S3."""
 
 
+class FileNotFoundError(AwsError):
```
betatim: I think we shouldn't inherit from `AwsError` here as the exception is only used for local files.

betatim:
Thanks for working on this! I left some feedback.

It would be super nice to add a test to make sure that the correct exception is raised when the directory does not exist and not raised when the directory is missing.

papermill/iorw.py
harsham4026: @betatim Sure. Will make the change as you suggested

papermill/exceptions.py
harsham4026: @betatim ok. Can I throw the normal Exception with the message that directory doesn't exists?

harsham4026:
@betatim Made the changes you suggested using the standard OS library for getting the directory path.

betatim:
I should have been clearer with my comment about the exception: I would keep the custom exception but I would not have it inherit from `AwsException` and instead inherit from `Exception`.

papermill/iorw.py
```diff
@@ -88,6 +88,11 @@ def listdir(cls, path):
 
     @classmethod
     def write(cls, buf, path):
+        dir = os.path.dirname(path)
+        if not dir.endswith("/"):
+            dir += "/"
+        if not os.path.exists(dir):
+            raise Exception('output folder {} doesn\'t exist!'.format(dir))
```
betatim: I'd rewrite this block as:
```python
dir_name = os.path.dirname(path)
if not os.path.exists(dir_name):
    raise NoSuchDirectory("The output folder {} does not exist.".format(dir_name))
```

and define a `NoSuchDirectory` exception

harsham4026: @betatim have you added this NoSuchDirectory exception or shall I add and commit the code?

harsham4026: @betatim added the code block and committed.

harsham4026:
@betatim yes I added a custom exception and inherited from Exception instead of AwsException.

papermill/exceptions.py
MSeal: @harsham4026 There's already a built-in FileNotFoundError in Python. Let's use that here :)

papermill/iorw.py
```diff
@@ -88,6 +88,11 @@ def listdir(cls, path):
 
     @classmethod
     def write(cls, buf, path):
+        dir = os.path.dirname(path)
+        if not dir.endswith("/"):
+            dir += "/"
+        if not os.path.exists(dir):
+            raise NoSuchDirectory('output folder {} doesn\'t exist.'.format(dir))
```
MSeal: Let's still use FileNotFoundError instead of making a new error class, as it's used for missing directories as well.

From Python docs:
| exception FileNotFoundError
| Raised when a file or directory is requested but doesn’t exist. Corresponds to errno ENOENT.

papermill/iorw.py
```diff
@@ -88,6 +88,11 @@ def listdir(cls, path):
 
     @classmethod
     def write(cls, buf, path):
+        dir = os.path.dirname(path)
+        if not dir.endswith("/"):
```
MSeal: I don't think you need this

papermill/iorw.py
```diff
@@ -88,6 +88,11 @@ def listdir(cls, path):
 
     @classmethod
     def write(cls, buf, path):
+        dir = os.path.dirname(path)
```
MSeal: Since `dir` is a reserved keyword in python let's use a different name here, or just combine this with the last line to get `if not os.path.exists(os.path.dirname(path)):`

harsham4026: @MSeal removed the dir and used os.path.dir(path)

papermill/iorw.py
harsham4026: @MSeal removed this piece of code

papermill/iorw.py
harsham4026: @MSeal used FileNotFoundError for throwing the error if a folder or file doesn't exist but the build is failing because FileNotFoundError is not directly supported in Python 2.7. Could you please let me know if we can proceed this way.

harsham4026:
@MSeal made the changes you suggested but the build is failing as **FileNotFoundError** directly not available to use in Python 2.7
screen shot 1940-06-22 at 12 30 33 am

harsham4026:
@MSeal added the test case to check if it throws error when we give a folder which doesn't exist and it's passed. Could you please consider having a review and merging it?

MSeal:
@harsham4026 You can add
```
try:
    FileNotFoundError
except NameError:
    FileNotFoundError = IOError
```
in both the iorw.py and the test file to fix the test error for python 2 (this is the recommended pattern as `FileNotFoundError` is a child of `IOError` and `IOError` is what python 2 built-in libraries use for this issue. Sorry I forgot python 2 didn't have this error type when I mentioned it before.

Issue #200, opened 2018-09-06
Further separate IO and notebook execution so the python API can operate on nbformat objects


VincentBLortie:
The current python API allows the user to execute a notebook by providing paths to input and output files, and potentially a path to a parameters file, through the `execute_notebook` function of `execute.py`.

Unfortunately, there doesn't seem to be a way to call papermill directly with nbformat notebook objects.

I'd like your thoughts on this suggestion:

- The calls to `load_notebook_node` and `write_ipynb` could be moved out of `execute_notebook`
- Alternatively, everything between these two calls could be extracted into a new public method that takes in and returns nbformat objects
- In either of these cases, the following lines should probably be extracted from the top of `_parameterize_notebook` and run at the same time as the rest of the IO, so that `_parameterize_notebook` is guaranteed to receive a dictionary:

```python
    # Load from a file if 'parameters' is a string.
    if isinstance(parameters, six.string_types):
        parameters = read_yaml_file(parameters)
```

Another option, which keeps the calls to iorw in execute_notebook but opens doors for users:

- input may be either a path or an object. If it is a path, then an object is loaded using load_notebook_node
- output may be either a path or None. An output file is only written if the path is not None
- return the resulting nbformat object in all cases
- nothing changes for the parameters

MSeal:
I think the logic would better fit with the #177 Issue around adding engine abstractions. The NBConvert engine could have an interface for taking it's internal object and parameterizing it.

Though in reality you're just applying parameterization to a notebook object. I'd rather not expose a lot of internal details of NBConvert in papermill, so maybe we can think about engines optonaoly exposing their representation layer, or being able to translate papermills's abstraction (in the case of NBConvert model today it would mostly return the same object type).

I do like the idea of allowing the iowr or execute layers to be a little more flexible to passing a model along rather than just a filename.

Do you have thoughts based on this line of thinking that you see fitting the desired goal?

Also I will be disconnected from the internet for several days, so likely won't respond until next week.

VincentBLortie:
I am a bit out of my depth here, since I haven't explored much of the codebase of nbconvert and other related projects.

To clarify, when you say you'd rather not expose too many internal details of NBConvert, you're referring to the nbformat representation? It's what `execute_notebook` returns, and I assumed (perhaps incorrectly) that it was somewhat of a standard for representing Jupyter notebooks.

No worries about being disconnected, I don't have much time tonight to think about this anyway.

VincentBLortie:
[...]

Issue #177, opened 2018-08-26
Add `--engine` command-line argument


betatim:
Right now notebooks can only be executed locally using nbconvert.

It would be nice to be able for the user to pick an "engine". For example to run notebooks on [kaggle via their API](https://github.com/Kaggle/kaggle-api) or on a BinderHub. It would also let you swap the current nbconvert engine for a xeus(?) based local runner.

First step would be to introduce the command-lien argument with it defaulting to "nbconvert". Preserving the current behaviour. Then add a binder or kaggle engine and exposing it.

MSeal:
:+1:

MSeal:
Implementation here would need to make an engine register class following translators / iorw files. Will be interesting to see what generic methods we'd want to support the first two engines.

betatim:
Another (speculative) use of this would be an engine that inspects the notebooks metadata to find out what dependencies the notebook has and then creates an environment (docker, virtual env, conda env) to run the notebook in on demand.

Issue #165, opened 2018-08-02
Notebook-friendly progress bars


SynapticSage:
Would it be useful to let users switch tqdm() to the module's tqdm_notebook() function? It works exactly the same, but prints progress bars better in notebooks, without forcing new progress messages onto a new line. Perhaps an interface could be exposed for users to set that function instance?

betatim:
Two ideas:

1. do people who run `papermill -p blah foo input.ipynb out/notebook.ipynb` on the command-line actually like seeing a progressbar? For me it is Ok that it is there but I don't particularly care for it. If no one loves it maybe we only have a progress-bar when called from Python?

1. allow users to choose the kind of progress-bar when they call papermill. The default if you are using papermill as a Python library could be "notebook style" and when it is called via the command line "terminal style" is the default.

I'd avoid trying to guess, that tends to go wrong too often, at which point you need the option for the user to specify it anyway :-/

MSeal:
1) has definitely been useful for users I've worked with. It gives initial trust that the tool is working and that it's not just stuck before it starts actual work.

2) Engines PR just merged which tries to be smart about it (so far in testing it works as expected in all environments). I think we could easily extend it to allow for user assignment if that makes sense for someone.

MSeal:
Since this issue got resolved with the merge of the engine PR, are we ok with closing this issue? We could open a new one if there's discussion of how to parameterize or change progress bar behavior elsewhere.

betatim:
I'm good with closing.

huynhducloi00:
the progress bar is printed on every line using papermill. this issue is NOT fixed.

PR #204, opened 2018-09-13
Papermill Engines

MSeal:
Attempts to implement #177 (and possibly #200) by adding an engine abstraction which allows for multiple executor options for running the notebook.


MSeal:
Still needs new tests for new functions, only fixed existing tests.

MSeal:
Was trying to refactor how we pass information to nbconvert, as I imagine future engines may implement `execute_cell` and have `execute_notebook` simply iterate through cell execution and apply the same `timed` wrapper found in the engine base class. But with nbconvert it's a tangle of side-effects in the Preprocessor. Could merge without fixing that and come back to the Preprocessor changes in an nbconvert upgrade PR instead(?).

papermill/engines.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class NBConvertEngine(EngineBase):
+    @classmethod
+    def execute_notebook(
+        cls,
+        nb,
+        kernel_name,
+        output_path=None,
+        progress_bar=True,
```
betatim: Was wondering if you want to leave the progress bar to each engine or have it as something "external".

I might have a slight preference for making the progress bar something that is provided (or not) by the caller of the engine. The engine gets passed a callback that it can (or not) call every time it makes progress. Something along the lines of https://github.com/tqdm/tqdm#manual

Maybe something for a new PR as it would require surgery on the executor.

MSeal: I think I generally agree with this idea. Let me play with it some here. Worse case we can do a second pass PR for items like this.

MSeal:
@betatim Take a look at this pass, still needs docs, new tests, and some better names... but now all the papermill abstractions around setting metadata, logging, and progress bars are captured in the `EngineNotebookWrapper` we pass to the `execute_notebook`. New engine clients only need to call `cell_start`, `cell_complete`, and `cell_exception` at the appropriate location in the engine subclass.

MSeal:
This also eliminates a lot of the custom junk we had added in the nbconvert preprocessor clone.

papermill/engines.py
```diff
@@ -0,0 +1,224 @@
...
+        processor = PapermillExecutePreprocessor(
+            timeout=execution_timeout,
+            startup_timeout=start_timeout,
+            kernel_name=kernel_name
+        )
+        processor.log_output = log_output
+        processor.preprocess(engine_nb, {})
+
+
+# Instantiate a PapermillIO instance and register Handlers.
```
betatim: copy&pasta "PapermillIO instance"

papermill/engines.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def set_pbar(self):
+        bar_format = self.DEFAULT_BAR_FORMAT
+        if self.log_output:
+            # We want to inject newlines if we're printing content between enumerations
+            bar_format += "\n"
+
+        self.pbar = tqdm(
+            total=len(self.nb.cells),
+            bar_format=self.DEFAULT_BAR_FORMAT)
```
betatim: -> `bar_format` (I think)

papermill/engines.py
```diff
@@ -0,0 +1,224 @@
...
+
+    no_tqdm = False
+except OSError:
+    no_tqdm = True
+
+from .preprocess import PapermillExecutePreprocessor
+from .iorw import write_ipynb
+
+
+def catch_nb_assignment(func):
```
betatim: could you add a comment about when/why this is needed? I think I understand what the code does but I am not sure I grok why it is needed or when it is needed.

betatim:
Do the methods in `EngineNotebookWrapper` get called by the nb convert processor? Should it be part of the official way a notebook gets executed?

Below is a snippet of code I use to execute notebooks on a remote machine via kaggle kernels. I think a binder engine would look similar in the level of detail/hooks:
```python
def run():
    # argument parsing and stuff
    argv = sys.argv[1:]
    args = get_runparser().parse_args(argv)
    kernel_dir = os.path.dirname(args.kernel)

    # take a notebook from a directory with some extra magic files in it
    # and send it to the remote host
    kaggle.api.kernels_push_cli(kernel_dir)

    metadata = get_metadata(kernel_dir)
    # wait for the kernel to fiish executing
    wait_for(metadata['id'])
    # fetch and write to the `outputs` subdirectory
    kaggle.api.kernels_output(metadata['id'], os.path.join(kernel_dir, 'outputs'))
```

The main point is that some of the engines I had in mind provide a lot less points to "hook" into the execution. We don't have to solve it now but I thought I'd share it so we can start pondering how that would fit while we build the first engine.

papermill/engines.py
MSeal: Whoops

MSeal:
I think we should accommodate up to the number of hooks presented. If the engine doesn't have cell level hooks it should still function. I'll make sure to add tests for a fake engine with fewer hooks.

papermill/engines.py
MSeal: Definitely. I'll do a comments and tests pass next

papermill/engines.py
MSeal: Good catch

MSeal: Good catch

papermill/cli.py
```diff
@@ -19,10 +19,11 @@
 def print_papermill_version(ctx, param, value):
     if not value:
         return
-    print("{version} from {path} ({pyver})".format(
-        version=papermill_version,
-        path=__file__,
-        pyver=platform.python_version()))
+    print(
```
MSeal: `black` changed this

papermill/engines.py
```diff
@@ -0,0 +1,310 @@
...
+        Initializes the progress bar object for this notebook run.
+
+        This is called automatically when constructed.
+        """
+        bar_format = self.DEFAULT_BAR_FORMAT
+        if self.log_output:
+            # We want to inject newlines if we're printing content between enumerations
+            bar_format += "\n"
+
+        if 'ipykernel' in sys.modules:
```
MSeal: This solves https://github.com/nteract/papermill/issues/165 too (was moving this and remembered to check if this was easy)

MSeal:
Will add more tests later this week, just did a little cleanup, commenting, and the first few tests on the plane.

MSeal:
@betatim What do you think about the comment from https://github.com/nteract/papermill/issues/220 ? Should we choose this point to adopt papermill API notebooks as the internal store wrapping nbformat objects or consolidate around nbformat only?

papermill/preprocess.py
```diff
@@ -49,10 +33,9 @@ class PapermillExecutePreprocessor(ExecutePreprocessor):
     # Copyright (c) IPython Development Team.
     # Distributed under the terms of the Modified BSD License.
 
-    # TODO: Delete this wrapper when nbconvert allows for setting preprocessor
-    # hood in a more convienent manner
+    # TODO: Delete this wrapper and push shared changes to nbconvert
```
rgbkrk: In order to make TODO bot do the issue tracking work for you in a nice clean way, switch from use of the word `this` and use something descriptive. "Delete the preprocess wrapper from PapermillExecutePreprocessor and push shared changes to nbconvert".

papermill/preprocess.py
```diff
@@ -138,68 +123,26 @@ def papermill_preprocess(self, nb, resources):
...
+            try:
+                engine_nb.cell_start(cell)
+                if not cell.source:
+                    continue
+                nb.cells[index], resources = self.preprocess_cell(cell, resources, index)
+            except CellExecutionError as ex:
+                engine_nb.cell_exception(nb.cells[index], exception=ex)
+                break
+            finally:
+                engine_nb.cell_complete(nb.cells[index])
```
rgbkrk: This is super clean.

papermill/engines.py
```diff
@@ -0,0 +1,328 @@
...
+            timeout=execution_timeout, startup_timeout=start_timeout, kernel_name=kernel_name
+        )
+        processor.log_output = log_output
+        processor.preprocess(engine_nb, kwargs)
+
+
+# Instantiate a PapermillEngines instance and register Handlers.
+papermill_engines = PapermillEngines()
+papermill_engines.register(None, NBConvertEngine)
+papermill_engines.register('nbconvert', NBConvertEngine)
```
rgbkrk: I'm definitely liking this overall structure.

papermill/preprocess.py
```diff
@@ -49,10 +33,9 @@ class PapermillExecutePreprocessor(ExecutePreprocessor):
     # Copyright (c) IPython Development Team.
     # Distributed under the terms of the Modified BSD License.
 
-    # TODO: Delete this wrapper when nbconvert allows for setting preprocessor
-    # hood in a more convienent manner
+    # TODO: Delete this wrapper and push shared changes to nbconvert
 
-    def preprocess(self, nb, resources):
+    def preprocess(self, engine_nb, resources):
```
VincentBLortie: Perhaps preprocess.py could be moved to an nbconvert directory or renamed to make it more obvious that the term "preprocessor" is to be understood in the context of nbconvert. Without prior exposure to nbconvert, I spent a while wondering what the notebook was being preprocessed for.

papermill/engines.py
```diff
@@ -0,0 +1,328 @@
...
+    RUNNING = "running"
+    COMPLETED = "completed"
+    FAILED = "failed"
+
+    def __init__(self, nb, output_path=None, log_output=False, progress_bar=True):
+        # Deep copy the input to isolate the object being executed against
+        self.nb = copy.deepcopy(nb)
+        self.output_path = output_path
+        self.log_output = log_output
+        self.set_timer()
```
VincentBLortie: `set_timer` is called in `__init__` and in `notebook_start`, which, in the base implementation of `wrap_and_execute_notebook`, occur one after the other. Perhaps this call isn't necessary?

papermill/engines.py
```diff
@@ -0,0 +1,328 @@
...
+
+        Args:
+            nb (NotebookNode): Executable notebook object.
+            kernel_name (str): Name of kernel to execute the notebook against.
+            log_output (bool): Flag for whether or not to write notebook output to stderr.
+            start_timeout (int): Duration to wait for kernel start-up.
+            execution_timeout (int): Duration to wait before failing execution (default: never).
+        """
+
+        processor = PapermillExecutePreprocessor(
```
VincentBLortie: `preprocessor`?

papermill/engines.py
```diff
@@ -0,0 +1,328 @@
...
+
+class EngineBase(object):
+    """
+    Base class from which engines should inherit and implement execute_notebook.
+    Defines wrap_and_execute_notebook method which is used to correctly setup
+    the EngineNotebookWrapper object for engines to interact against.
+    """
+
+    @classmethod
+    def wrap_and_execute_notebook(
```
betatim: Can we swap the names of `wrap_and_execute_notebook` and `execute_notebook`?

Somehow `execute_notebook` is the thing I was expecting to interact with as a user of an `Engine` instance. But now I think actually I'd be using `wrap_and_execute_notebook`.

I propose `execute_notebook` (new name of `wrap_and_execute_notebook) as the entry point for users, and `execute_wrapped_notebook` (new name for `execute_notebook`) as the thing engine implementors have to implement.

Just feels weird to be using the function with the complex name and as a user "I don't want my notebook wrapped, just executed please"

WDYT?

papermill/engines.py
```diff
@@ -0,0 +1,328 @@
...
+    def cleanup_pbar(self):
+        if hasattr(self, 'pbar') and self.pbar:
+            self.pbar.close()
+            self.pbar = None
+
+    def __del__(self):
+        self.cleanup_pbar()
+
+
+class EngineBase(object):
```
betatim: Calibration question: `Engine` seems like a much nicer name than `EngineBase`, am I the only one who is bothered by this? I can't rationally explain why it bothers me, it just does, like a picture that isn't hanging perfectly straight ... :)

papermill/engines.py
```diff
@@ -0,0 +1,328 @@
...
+        nb = kwargs.get('nb')
+        if nb:
+            # Reassign if executing notebook object was replaced
+            self.nb = nb
+        return func(self, *args, **kwargs)
+
+    return wrapper
+
+
+class EngineNotebookWrapper(object):
```
betatim: It took me several attempts to get what this piece does and how it is used. Now I think of it as thing that hold metadata and state about the execution progress of a notebook. It provides some hooks I can call to update the state as I (the engine) churns through the notebook.

I don't know if a different name would have helped me get this idea sooner or not. Some suggestions: `NotebookExecutionState` or `EngineProgressState`? Wrapper makes me think we wrap some layer around the notebook to make it compatible with how an engine expects it to look.

betatim:
I like it! It also felt like several 💡s went on in my head this time around reading this PR. As a result most comments are about naming because I feel that could have helped me grok it earlier??

papermill/engines.py
MSeal: I've typically used `Base` as suffix for abstract classes. I'm fine with changing it to `Engine` though.

MSeal:
Naming suggestions welcome, I just picked names without too much thought initially.

papermill/preprocess.py
MSeal: Yea the preprocessor name is really misleading in nbconvert. I can isolate the nbconvert names a bit more as you suggested.

papermill/engines.py
MSeal: I used this call to lazily set the `start_time` and `end_time`, I'll change it to some `= None` assignments to be clearer

papermill/engines.py
MSeal: I was struggling with this class name. Here's a few more ideas `EngineExecutionTracker`, `EngineProgressor`, `EngineExecutionState`? I might be leaning towards your suggestion of `EngineProgressState`. Anyone else have suggestions / think one of these sounds good.

papermill/engines.py
MSeal: That's a good point. I maybe optimized the naming for engine developers as they just needed to focus on `execute_notebook` implementation. I'm good with potentially swapping the names the way suggested here.

papermill/engines.py
betatim: One more `EngineState`? (I actually like `EngineProgressState` the least, progress and state are too similar, feels like `StripedZebra` :-/)

mpacer: I don't think progress and state are too similar — progress suggests tracking where it is in it's trajectory, state merely indicates what the current state is.

papermill/engines.py
mpacer: This isn't an abstract class though, since it's inheriting from object. If you were to want to make it an actual abstract class, one solution I've liked elsewhere in the Jupyter ecosystem is appending `ABC` to the end of the name.

papermill/engines.py
VincentBLortie: Remember that it's through this class that the engine can access the notebook. I imagine that's at least as important to mention in naming the class as the progress tracking part?

It's a notebook viewed through progress-tracking-tinted glasses. I can't think of a name that doesn't sound silly... A `ProgressingNotebook`?

MSeal:
Naming changes made -- I think the new names are clearer and I took the suggestions where we had consensus.

I also took the opportunity to cleanup the nbconvert preprocessor wrapper (run_cell can't be wrapped/removed yet without some tweaks upstream). It now uses the logger object for messaging and has the nbconvert 5.4 common code removed from papermill.

MSeal:
I feel good with a merge. I'll hold off doing a release for a little bit in case there's any syntax requests or issues discovered after merge. (I did a bunch of testing last night outside of the unittests too)

PR #209, opened 2018-09-16
changes for displaying the version of papermill

harsham4026:


harsham4026:
@MSeal I made couple of changes for --version but the test cases are giving some error. I am not understanding those errors. Please help me put to fix these errors.

papermill/cli.py
```diff
@@ -10,10 +10,18 @@
 
 from .execute import execute_notebook
 from .iorw import read_yaml_file
+from .api import Notebook as notebook
```
MSeal: So the intention is to print the papermill library version rather than the notebook version.

```from . import __version__ as papermill_version```

will get you the version of the library being used. What you have here will print `<property object at 0x7f54aa88baf8>` instead :)

File: papermill/cli.py [hunk 2]
```diff
@@ -50,6 +58,8 @@
     '--start_timeout', type=int, default=60, help="Time in seconds to wait for kernel to start."
 )
 @click.option('--report-mode/--not-report-mode', default=False, help="Flag for hiding input.")
+@click.option('--version', default=1.0, is_flag=True, callback=print_the_paermill_version,
```
MSeal: You don't want the default section here.

papermill/cli.py
```diff
@@ -10,10 +10,18 @@
 
 from .execute import execute_notebook
 from .iorw import read_yaml_file
+from .api import Notebook as notebook
 
 click.disable_unicode_literals_warning = True
 
 
+def print_the_paermill_version(ctx, param, value):
```
MSeal: `print_papermill_version`

papermill/cli.py
```diff
@@ -95,6 +106,7 @@ def papermill(
         log_output=log_output,
         start_timeout=start_timeout,
         report_mode=report_mode,
+        version=version,
```
MSeal: Don't want or need to pass this down.

papermill/execute.py
```diff
@@ -25,6 +25,7 @@ def execute_notebook(
     log_output=False,
     start_timeout=60,
     report_mode=False,
+    version=1.0,
```
MSeal: Skip, don't need past the CLI

papermill/tests/test_cli.py
```diff
@@ -357,6 +373,23 @@ def test_no_report_mode(self, execute_patch):
             progress_bar=True,
             start_timeout=60,
             report_mode=False,
+            version=1.0,
+        )
+
+    @patch(cli.__name__ + '.execute_notebook')
+    def test_version(self, execute_patch):
+        self.runner.invoke(papermill, self.default_args + ['--version'])
```
MSeal: drop `default_args`

papermill/tests/test_cli.py
```diff
@@ -357,6 +373,23 @@ def test_no_report_mode(self, execute_patch):
...
+        execute_patch.assert_called_with(
```
MSeal: If we do this right, `execute_notebook` should not be called with `--version`. Test here should say `execute_patch.assert_not_called()` when it's working.

papermill/cli.py
harsham4026: removed

papermill/execute.py
harsham4026: removed

papermill/cli.py
harsham4026: made the change

papermill/cli.py
harsham4026: done!

harsham4026:
@MSeal made all the changes and the test cases passed as well. Please consider this merging and let me know if any changes need to be made.

papermill/execute.py
```diff
@@ -37,6 +37,7 @@ def execute_notebook(
         log_output (bool): Flag for whether or not to write notebook output to stderr.
         start_timeout (int): Duration to wait for kernel start-up.
         report_mode (bool): Flag for whether or not to hide input.
+        version : Flag for knowing the version of the papermill running.
```
MSeal: Missed undoing this comment :)

papermill/cli.py
MSeal: I meant we should rename the function :)

harsham4026:
@MSeal undone the comments regarding the version.

papermill/cli.py
```diff
@@ -10,10 +10,18 @@
...
 from .iorw import read_yaml_file
+from . import __version__ as papermill_version
 
 click.disable_unicode_literals_warning = True
 
 
+def print_the_paermill_version(ctx, param, value):
+    if not value:
+        return
+    print(papermill_version)
```
MSeal: Let's make this slightly more verbose (like how pip --version works)

> $ pip --version
> pip 18.0 from /home/mseal/.py2local/local/lib/python2.7/site-packages/pip (python 2.7)

```
    print("{version} from {path} ({pyver})".format(
          version=papermill_version,
          path=__file__,
          pyver=platform.python_version()))
```
should do the trick

harsham4026: @MSeal Done!!

papermill/cli.py
harsham4026: @MSeal can you please suggest any name for changing the name of function?

harsham4026:
@MSeal I think almost all the changes have been made suggested by you :)

papermill/cli.py
harsham4026: @MSeal made the change and committed. Now it's looking cool.

MSeal:
Thanks for all the PR's. We appreciate the contributions and hopefully the coding feedback is helpful for you as well :)

harsham4026:
@MSeal yes definitely the feedback is helping me in improving my coding standards :)

PR #212, opened 2018-09-20
 Using logger instead of print statement

schrockn:
Currently execute_notebook is hard-coded to print its two messages
to stderr using print. This PR changes this to use the python
logging framework.

First PR so let me know if you want this at all or in a totally different way. Thanks!

P.S. This project is so cool. Thank you for open sourcing it.

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -3,8 +3,8 @@
 from __future__ import unicode_literals, print_function
 
 import datetime
+import logging
 import os
-import sys
 import six
 import nbformat
 
@@ -13,6 +13,9 @@ from .exceptions import PapermillExecutionError
 from .preprocess import PapermillExecutePreprocessor
 from .iorw import load_notebook_node, write_ipynb, read_yaml_file, get_pretty_path
 from .translators import translate_parameters
+
+logging.basicConfig(level=logging.INFO, format="%(message)s")
+logger = logging.getLogger('papermill')
 
 
 def execute_notebook(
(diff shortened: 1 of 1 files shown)


MSeal:
Hi @schrockn ! Thanks for opening the PR (and for the project praise).

I do agree we should control logging here, but I think we can capture it a little better than we do today. Instead of setting a `printer` method I think we should pass a [logger object](https://docs.python.org/3/library/logging.html) which defaults to something like [this post](https://stackoverflow.com/questions/14058453/making-python-loggers-output-all-messages-to-stdout-in-addition-to-log) except with stderr as the output channel. This way we conform to a more known interface and allow users to alter the logging pattern with known conventions easier.

schrockn:
@MSeal do you want this to default to just the root logger? Or do you want a named logger with the `papermill` namespace?

```
import logging
logging.info('message')
```
or

```
import logging
logger = logging.getLogger('papermill')
logger.info('message')
```

MSeal:
Actually, @betatim pointed out to me that we should instead just reference the logger rather than pass it around. We could let the execute function either pass the log level (which we could expose to the CLI layer easily) or external code could load the named logger.

schrockn:
Updated per-request. Python logging framework is awkward to test without library support so I have removed the test testing the actual messages..(LogCapture in https://github.com/Simplistix/testfixtures/ is promising but adding that as a dependency is it's own discussion)

PR #222, opened 2018-09-28
Adding the name to output string

choldgraf:
I noticed that a string isn't being properly populated w/ its corresponding value!

### DIFF
--- a/papermill/tests/test_api.py
+++ b/papermill/tests/test_api.py
@@ -1,4 +1,5 @@
 import unittest
+import pytest
 
 try:
     from unittest.mock import patch
@@ -35,6 +36,10 @@ class TestNotebookClass(unittest.TestCase):
             columns=['name', 'value', 'type', 'filename'],
         )
         assert_frame_equal(nb.dataframe, expected_df)
+
+        with pytest.raises(PapermillException) as excinfo:
+            nb.display_output('foo')
+        assert "Output Name 'foo' is not available in this notebook." in str(excinfo.value)
 
     def test_bad_file_ext(self):
 
--- a/papermill/api.py
+++ b/papermill/api.py
@@ -216,7 +216,7 @@ class Notebook(object):
         """
         outputs = _get_notebook_outputs(self.node)
         if name not in outputs:
-            raise PapermillException("Output Name '%s' is not available in this notebook.")
+            raise PapermillException("Output Name '%s' is not available in this notebook." % name)
         output = outputs[name]
         ip_display(output.data, metadata=output.metadata, raw=True)
 


betatim:
Nice work!

Could you also add a test to make sure this doesn't come back in the future? Maybe call `display_output` with a mocked `_get_notebook_outputs` checking that it raises/doesn't raise an exception.

choldgraf:
hmmm - so I'm not sure the best way to test this. The exception itself is being raised fine, it is the *text* of the exception that's wrong. Anyone know of a way to check the text of an exception in tests (w/o building in a lot of complexity?)

MSeal:
Something like

```
with self.assertRaisesMessage(PapermillException, '"Output Name foo is not available in this notebook."'):
  nb.display_output('foo')
```

betatim:
Another way to do the same (for the pytest enthusiasts) https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

```python
with pytest.raises(PapermillException) as excinfo:
    nb.display_output('foo')
assert 'Output Name foo is not available in this notebook.' in str(excinfo.value)
```

choldgraf:
ok I think I figured it out - it was sort of a hybrid between both of your suggestions :-)

I didn't find a `assertRaisesMessage` method for the `unittest.TestCase` subclass, and it also didn't seem like pytest was being used so I didn't use that. In the end, it looks like `excinfo.exception` was the magical incantation

papermill/tests/test_api.py
```diff
@@ -36,6 +36,10 @@ def test(self):
         )
         assert_frame_equal(nb.dataframe, expected_df)
 
+        with self.assertRaises(PapermillException) as excinfo:
```
mpacer: `pytest.raises` rather than self.assertRaises?

papermill/tests/test_api.py
```diff
@@ -36,6 +36,10 @@ def test(self):
...
+            nb.display_output('foo')
+        assert "Output Name 'foo' is not available in this notebook." in str(excinfo.exception)
```
mpacer: Does excinfo.value not work per this [`pytest.raises`](https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions) example?

papermill/tests/test_api.py
MSeal: To be fair, this is a unittest object for self.Raises is perfectly valid.

MSeal:
No need to spend a ton of cycles on testing a 1 line change for a message. If you want to try the pytest version awesome, otherwise I'm ok merging now for such a trivial change and we can do a test update pass in the future for these.

choldgraf:
I'm happy to use pytest if it's better, I just didn't add to that test file since it wasn't being used anywhere else in there. Is papermill migrating away from unittest?

MSeal:
Yes, gradually. I am probably the user keeping it in unittest the most by
not learning pytest equivilents everywhere :P. But in general pytest is
preferred here.

betatim:
Travis is happy so I'm merging!

Thanks for the fix and humouring us with the testing extravaganza!

ps. the question for y'all to consider: who tests the tests?! 😱

choldgraf:
cool - the latest push swapped out pytest for unittest 👍

Issue #181, opened 2018-08-26
Add notebook example for running notebooks from other notebooks and managing error cases


MSeal:
Many people requested examples -- perhaps we can upgrade binder to show this pattern well.

PR #223, opened 2018-09-28
adding example for binder

choldgraf:
I was playing around with the papermill API to learn how it works, and I decided to write up a little example notebook to help myself learn. I thought it might be useful for others to learn how to use papermill as well, so here's a PR that:

1. Sets up a Binder folder so we can install some requirements needed for an example
2. Adds an example notebook that demonstrates a (sort-of) "real world" usecase for papermill, and highlights some of its functionality.
3. Updates binder links to point to this example.

Open to suggestions/comments/etc!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,3 +105,6 @@ ENV/
 
 # pytest
 .pytest_cache
+
+# Binder example
+binder/*run*.ipynb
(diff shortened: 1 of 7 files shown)


MSeal:
Hey that's really awesome! Thanks for putting this together. Can you add (or explicitly exclude) the files from the manifest check so Travis is happy?

I think this partially solves https://github.com/nteract/papermill/issues/181 too. I might add another notebook alongside this and reference both in the readthedocs.

choldgraf:
ok travis should be happy now!

PR #224, opened 2018-10-01
Pyarrow hdfs support

harsham4026:


harsham4026:
@MSeal made the changes for supporting the hdfs using pyarrow client. Please review the PR.

harsham4026:
@MSeal this build is getting failed for all the Python versions except Python 3.5. I added Cython in dependency but it's failing because of Cython and pyarrow's compat. Any suggestions to resolve this please?

MSeal:
I might not get to looking at this until tomorrow. Two things to note, one we should avoid adding cython as a dependency if we can (complicates our build options heavily). Second this github issue seems to have some information on pyarrow in 3.7 builds: https://github.com/apache/arrow/issues/1125

MSeal:
Posted an issue on pyarrow about this: https://github.com/apache/arrow/issues/2708

We might need to restrict the feature to python 3.6 or less, or include it in a side package. I don't want cython and a bunch of build dependencies added if the wheel doesn't work out-of-the box.

xhochy:
We will release wheels for `pyarrow` for all platforms. These then don't require `Cython`. As we hope to ship a new `pyarrow` release the next days with Python 3.7 wheels, I would suggest to just to review the code here and wait with the merge until `pyarrow==0.11` is released. Then there is no need to restrict to Python < 3.7.

MSeal:
Thanks @xhochy , appreciate the input and quick response!

rgbkrk:
This PR has been open long enough that I'm going to recommend trying it again as a new PR, given how much has changed since October. Feel free to mention this PR in any new PRs related to this one.

Cheers!

PR #229, opened 2018-10-09
[WIP] First attempt at making a kaggle backend

betatim:
A quick exploration of using kaggle.com kernels as backend/engine for papermill.

After setting yourself up locally to use [the kaggle api](https://github.com/Kaggle/kaggle-api) try this with:
```
papermill --engine kaggle blahaha/some-notebook.ipynb /tmp/blah.ipynb
```

This will create a new kernel on kaggle.com and execute it. However currently you can't retrieve the executed notebook via the API so we [can't get the output back](https://twitter.com/rctatman/status/1049358398641205248).

You could imagine doing something completely hacky by inserting a cell at the end of the notebook that tries to find the notebook server, ask it for the contents of the notebook, and then write it to the "outputs" of the kaggle kernel. More productive: wait for kaggle to add this feature.

### DIFF
--- a/papermill/engines.py
+++ b/papermill/engines.py
@@ -1,7 +1,11 @@
+import os
 import sys
 import copy
+import json
+import time
 import datetime
 import dateutil
+import tempfile
 
 from functools import wraps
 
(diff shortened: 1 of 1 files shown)


betatim:
One thing I did find with making this change was that having `execute` be a class method felt weird. As in things that are unusual are weird. Mostly I wanted to use the class instance (`self.`) to store stuff computed in the constructor and then used when executing the notebook. It feels like now we are using classes as a way of namespacing functions, which we could do with namespaces.

In the end it all worked but I left a bit puzzled about what the thought process was behind things being this way. Maybe the actionable thing here is that we add a few comments to the engines code to explain the why (so I won't be puzzled again in three months time when I've forgotten).

MSeal:
I like your hack idea for nabbing the notebook (mostly to see if it would actually work). But I agree wait for feature is probably better.

The class method pattern I put in to be consistent with the other registered classes. I see how it enforces a functional way of representing a group of functions meant to run together, but I had the same reservation when I first saw the pattern with io.py. There's nothing saying we couldn't change the pattern.

willingc:
@betatim Should we leave this open or go ahead and close?

betatim:
If I remember correctly this stalled because [we can't retrieve the notebook](https://twitter.com/rctatman/status/1049358398641205248). I never followed up to check if this is now possible and my use-case has gone -> close.

MSeal:
As far as I can tell the feature wasn't added based on scanning issues / API docs. Closing this PR then

Issue #233, opened 2018-10-17
Update README/Docs to include a note writing to S3 when you have multiple accounts


aabdullah-bos:
__In a world__, where you need to run papermill across different accounts.  You can set the `AWS_PROFILE` environment variable. So that papermill uses the appropriate credentials to store the results of a notebook execution. I posted this in slack, and I am creating this issue so that it can be associated with a pull request to update the docs.

PR #234, opened 2018-10-17
Add note on using multiple AWS accounts

aabdullah-bos:
- If you set AWS_PROFILE, you can specify which
  credentials to use for storing results to s3.

Addresses issue #233


rgbkrk:
As mentioned in the nteract slack, we'll probably want to move this into `docs/usage.rst` so the README doesn't get so long that it's hard to get to a quick synopsis (we can always link to the docs from the README). This is probably small enough though that we can merge it as is and then go for another PR that updates both.

rgbkrk:
Thank you @aabdullah-bos!

Welcome to the nteract party! 🎉

PR #236, opened 2018-10-17
Fixed checks for ipython kernel to run before ipython might be imported by other functions

MSeal:
Noticed the import value changed after using nbformat function which rely on ipython. This fixes the issue by preloading the check.

### DIFF
--- a/papermill/engines.py
+++ b/papermill/engines.py
@@ -5,6 +5,11 @@ import dateutil
 
 from functools import wraps
 
+from .log import logger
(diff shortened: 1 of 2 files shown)


MSeal:
Before this passing parameters would cause the notebook to execute as though it were in a kernel and display the wrong progress bar type from command line.

betatim:
Restarted the builds as it seems unlikely this changed broke the jupyter messaging spec.

PR #237, opened 2018-10-18
stick an s in install

rgbkrk:
Since I copied `papermill` directly when creating `bookstore`, a missing `s` that @willingc found in https://github.com/nteract/bookstore/pull/1 was also present here. I have now typed more than 234x the characters that this PR introduces.


betatim:
Thanks for helping with the little details.

PR #242, opened 2018-10-31
Removed environment variable tracking for consumers

MSeal:
This was originally here at the very beginning to help with execution contexts. We never use the feature and it's definitely a security concern if secrets are sitting in env variables (or just system information).


betatim:
👍 

I think it isn't as bad as it looks at first sight as only whitelisted environment variables were stored.

We can also remove https://github.com/nteract/papermill/blob/master/papermill/conf.py now as it seems like env variables was the only use of it.

PR #244, opened 2018-11-04
Anchorage

mpacer:
This is a PR that should make it easier to navigate to the cell that was in error when you run into an error. Specifically, it allows you to click on a link that jumps you right to the cell where the error occurred.

How does it do this?

1. Give each cell a unique id
2. Create a new collection of cells that add anchors to the page tying that unique id to that location in the DOM
3. Modifies the error message being inserted at the top of the page to include an anchor link to the cell in error.

It does nothing if there is no error. 

I think I'd want to include tests for the new apis that this introduces before we released this, but I want to iterate a couple of times first on the exact form this should take. 

But nonetheless the existing tests needed to be modified to pass as of this latest commit.

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -163,36 +163,40 @@ ERROR_MESSAGE_TEMPLATE = (
 )
 
 ID_MSG_TEMPLATE = "<h1 id={cell_id}>Cell ID {cell_id}<a href=#{cell_id}>¶</a></h1>"
+
+
 def add_cell_id_metadata(cells):
     for cell in cells:
         cell.metadata["papermill"] = cell.metadata.get("papermill", {})
-        if not cell.metadata.get("papermill",{}).get("cell_id"):
-            cell.metadata['papermill'].update({"cell_id": "a" + str(uuid.uuid4())}) 
+        if not cell.metadata.get("papermill", {}).get("cell_id"):
+            cell.metadata['papermill'].update({"cell_id": "a" + str(uuid.uuid4())})
     return cells
 
+
 def html_cells_with_anchors(cells):
-    other_cells = copy.deepcopy(cells)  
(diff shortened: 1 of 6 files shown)


MSeal:
This is a neat idea. I like the functionality but I am a little hesitant to add all the extra cells. From a presentation aspect it looks mostly right (maybe hiding the cell ID output as well from display) but from a model pov we have many extra cells. Classic notebook would render this oddly today and the cell execution counts would (I believe?) all be even numbers with the markdown cells in-between.

Do you think maybe we could insert a hidden html output on each cell which includes the cell id in the same manner as this? That's eliminate the cell count / visualization issue but let you stlll jump to failure outputs?

I'll see if I can chat 1:1 with you to see if we can iterate on the approach and desired end-point offline

willingc:
I like the concept of this especially if the notebook has 20+ cells. Is there a way to suppress the output of the Cell IDs?

rgbkrk:
I'd be happy if the `cellID` was in the top level metadata and that we could use the same entry when writing notebooks with nteract. We could then acknowledge this formally by providing `id`s to the cell div when we render them with commuter.

willingc:
This looks pretty cool in the latest GIF. @rgbkrk @MSeal thoughts?

MSeal:
The user interface in commuter looks good now. I do still worry about the logic of doubling the number of cells, especially with not all interfaces respecting hidden fields yet.

Iterating on the cells for an output notebook now would be quite different than iterating on the input cells. Would it be possible for us to add the display output to the existing cells rather than make new cells? Each cell could have a hidden html output linking to the next cell as it's final output. Thoughts on taking that approach?

MSeal:
Closing for now. We can return to this topic again at a later point, or maybe try to get something more standardized elsewhere for generic support of this capability..

Issue #205, opened 2018-09-13
Add Azure Blob Storage backend


xhochy:
I would like to use `papermill` but store my notebooks on Azure Blob Storage. To achieve this, I think all it needs is implementing the same interface as done for S3 in `s3.py` and then registering it with `papermill_io.register("wasb://", WASBHandler)`?

MSeal:
Hi xhochy!

Yes that's all you have to add. You can follow the pattern in https://github.com/nteract/papermill/pull/184 which added Azure Data Lake Storage. The same libraries it used may also be usable for Azure Blob Storage (haven't checked).

MSeal:
Also https://github.com/nteract/papermill/blob/master/DEVELOPMENT_GUIDE.md#io-sources--sinks may be of use for how this works and what's expected with additions

PR #246, opened 2018-11-07
Add azure blob storage backend support

tim-werner:
Azure blob storage backend are made available using the abs:// prefix. 

Access will be granted using shared access signatures (SAS) (see https://docs.microsoft.com/en-us/azure/storage/common/storage-dotnet-shared-access-signature-part-1)

This pull request addresses the issue https://github.com/nteract/papermill/issues/205


MSeal:
Thanks for making the PR, it's nice and clean!

PR #249, opened 2018-11-11
Fixed botocore change break.

MSeal:
This needs to be merged asap as papermill is broken with recent boto versions. Tested with some basic s3 notebooks on python 2 and 3 beyond the unittests.

Also fixed the moto test breakage so we can unpin botocore and moto.

The lines touched outside of s3 are the result of black / flake passes cleaning up stray lines.

Took the opportunity to strip out all the unused junk in s3.py. This brings out test coverage up to 90% :confetti_ball:


willingc:
Assume you are cutting a dot release with this merge too.

MSeal:
Yep. Once I get travis to accept the pytest.ini change I'll get a release going.

rgbkrk:
Awesome, thank you @MSeal for the refactor + release. Thanks @willingc for the merge!

PR #250, opened 2018-11-12
Bugfix: AzureBlobStore.read returns a string but papermill expects a list of strings

tim-werner:
Papermill expects that the read function returns a list of strings, however `AzureBlobStore.read` returns a string and hence reading from azure blob storage fails with:

```
papermill "abs://myaccount.blob.core.windows.net/sascontainer/sasblob.txt?sastoken" hello_world.ipynb

[...]
    return "\n".join(lines)
TypeError: can only join an iterable
[...]
```

This pull request will fix this issue. Reading from azure blob storage was tested with Python2 and Python3

### DIFF
--- a/papermill/abs.py
+++ b/papermill/abs.py
@@ -1,4 +1,5 @@
 import re
+import io
 from six.moves import urllib
 from azure.storage.blob import BlockBlobService
 
@@ -32,12 +33,21 @@ class AzureBlobStore(object):
 
     def read(self, url):
         params = self._split_url(url)
+        output_stream = io.BytesIO()
         block_blob_service = self._block_blob_service(
             account_name=params["account"], sas_token=params["sas_token"]
         )
-        return block_blob_service.get_blob_to_text(
-            container_name=params["container"], blob_name=params["blob"]
+
+        block_blob_service.get_blob_to_stream(
+            container_name=params["container"],
+            blob_name=params["blob"],
+            stream=output_stream,
         )
+
+        output_stream.seek(0)
+        return [
+            line.decode("utf-8") for line in output_stream
+        ]
 
     def listdir(self, url):
         params = self._split_url(url)
--- a/papermill/tests/test_abs.py
+++ b/papermill/tests/test_abs.py
@@ -4,9 +4,21 @@ import six
 from ..abs import AzureBlobStore
 
 if six.PY3:
-    from unittest.mock import Mock
+    from unittest.mock import Mock, patch, ANY
 else:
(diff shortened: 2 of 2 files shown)


papermill/abs.py
```diff
@@ -32,13 +34,24 @@ def _split_url(self, url):
...
-            container_name=params["container"], blob_name=params["blob"]
+
+        block_blob_service.get_blob_to_stream(
+            container_name=params["container"],
+            blob_name=params["blob"],
+            stream=output_stream,
         )
 
+        output_stream.seek(0)
+        content = output_stream.readlines()
```
betatim: remove this if you like the next comment

papermill/abs.py
```diff
@@ -32,13 +34,24 @@ def _split_url(self, url):
...
+        return [
+            line.decode("utf-8") if isinstance(line, six.binary_type) else line
+            for line in content
```
betatim:
```suggestion
            for line in output_stream
```
does this work in Python 2 and 3?

papermill/abs.py
```diff
@@ -32,13 +34,24 @@ def _split_url(self, url):
...
+        block_blob_service.get_blob_to_stream(
+            container_name=params["container"],
+            blob_name=params["blob"],
+            stream=output_stream,
         )
 
+        output_stream.seek(0)
+        content = output_stream.readlines()
+        return [
+            line.decode("utf-8") if isinstance(line, six.binary_type) else line
```
betatim: Just for my education: when does it happen that some lines are `bytes` and others are not within the same stream?

tim-werner: You are right. Checking the lines to be `bytes` is unnecessary. The BytesIO stream should always return `bytes`. I will change this

papermill/abs.py
tim-werner: I like your suggestions. I will change this accordingly.

tim-werner: By the way. This works for Python 2 and 3

betatim:
LGTM. Don't have a local setup to test it either but I assume Tim W does :)

betatim:
> Of course it would be nice to have an integration test that tests all backends with a real storages behind it.

Getting that started sounds like a great next contribution :D

Maybe a less ambitious, first step is to collect what open-source implementations exist for the various backends. For S3 I know there is https://minio.io/ which is super easy to setup. Is there something for Azure?

tim-werner:
https://minio.io/ also supports azure. I will have a look

Issue #253, opened 2018-11-13
Hiding Ingested Parameters when executing with `--report-mode`


LeonardAukea:
I want to be able to hide the ingested parameters at least when running in report mode. Since a new cell is created by papermill when feeding in params, there is no way to add metadata for that cell in the notebook. When you want to execute a notebook in order to generate some sort of report where no code is visible I think that in most cases the ingested parameters should be hidden. #135 

I'm not super familiar with the codebase but would it be as simple as adding: 

```
newcell.metadata['jupyter']['source_hidden'] = True
``` 

Around here -> https://github.com/nteract/papermill/blob/master/papermill/execute.py#L112-L114 ?

----
Furthermore, it would be nice to have a neat way of handling secrets. According to my understanding it's currently not a very good idea to ingest secrets as parameters since, they are made available in the output notebook.

betatim:
That looks like it is the right place. Do you want to make a PR to add that functionality? I think it makes sense to add this metadata also for generated cells.

Secrets are tricky. Right now I'd use environment variables to make them available to the notebook. This isn't a absolutely, will always work solution though, see https://github.com/nteract/papermill/pull/232 for some discussion on the topic of setting environment variables for things like remote execution engines :-/ Ideas would be welcome, probably best in a new issue.

LeonardAukea:
Ok, I hope that small fix would do the trick #254

PR #254, opened 2018-11-13
Hiding Ingested Parameters if `--report-mode`

LeonardAukea:
Added functionality to hide ingested parameters in output notebook if executed with `--report-mode`

I don't think it makes sense to show the ingested parameters in the output notebook when using `--report-mode`.  Here is the issue I created. #253


MSeal:
Thanks for the fix. You have a trivial whitespace issue making flake8 complain. I can merge once you clean that line up.

LeonardAukea:
@MSeal Whitespace is removed.

MSeal:
Thanks for the fix!

wrathagom:
I think this PR may have introduced a small error as reported on #282

PR #259, opened 2018-11-30
Do not set logging level on import

skizunov:
When papermill is imported, the logging level was automatically being
set to `INFO`. Many Python libraries are chatty at that logging level
which leads to confusing messages when a user executes a Jupyter
notebook that imports `papermill` and also uses another Python library
to perform some task.

Logging will work correctly even if the call to `logging.basicConfig`
is omitted. It will simply log at the default logging level of the
environment.

Libraries in general should not change the logging level of the
environment at the import level, since this confuses users and is
difficult for users to understand what went wrong when they see
messages they don't expect nor, in many cases, understand.

Signed-off-by: Sergey Kizunov <sergey.kizunov@ni.com>

### DIFF
--- a/papermill/cli.py
+++ b/papermill/cli.py
@@ -3,6 +3,7 @@
 from __future__ import unicode_literals
 
 import base64
+import logging
 
 import click
 
@@ -95,6 +96,8 @@ def papermill(
     output in the destination notebook.
 
     """
+    logging.basicConfig(level=logging.INFO, format="%(message)s")
+
     if progress_bar is None:
         progress_bar = not log_output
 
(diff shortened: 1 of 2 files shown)


betatim:
Thanks for this change!

Should we change the logging level in [the entrypoint of the CLI](https://github.com/nteract/papermill/blob/5e2bda67e3aed8bbfc72cf17fc93081dab7afb59/papermill/cli.py#L75) instead to maintain the loglevel that people are used to when running papermill from the console? Or can we change the log level of just the papermill logger?

MSeal:
Let's move it into the CLI function. I think defaulting to logging config other tools use when importing makes sense, rather than forcing the user to also set the "papermill" logger settings everytime.

skizunov:
I moved `logging.basicConfig` into the CLI function.

betatim:
Nice!

Issue #257, opened 2018-11-19
Isolate IO dependencies out of default install


MSeal:
We're getting a longer list of possible destinations for papermill to read/write against. This is making the dependency graph grow in ways that will become problematic for lambda executions or small projects.

The idea is to leave all dependencies installed with papermill[all] but isolate to only the basic for normal installation paths.

TODO: Test for different install patterns in jenkins before declaring success

betatim:
Is it possible to have `pip install papermill[s3,aks]` to let people select multiple backends?

MSeal:
That would be ideal, we need to experiment with extras_requires to see what's possible

betatim:
Looks like multiple extras works! Checkout [example 6 of the docs](https://pip.pypa.io/en/latest/reference/pip_install/#examples)

PR #260, opened 2018-11-30
Allow custom handlers via entry points

janfreyberg:
This PR implements loading of IO handlers via entry points. This allows a third-party developer to implement a handler for papermill and declare it in `setup.py` under entry points.

The aim is to allow an ecosystem of papermill read / write options. This would allow, for example, to write custom papermill publishing methods that integrate with an organisation's infrastructure, or for a third-party hosting service (e.g. [Digital Ocean Spaces](https://www.digitalocean.com/products/spaces/) ). I have had a lot of success writing custom output formats for [RMarkdown](https://rmarkdown.rstudio.com/), and think allowing this for papermill would be a great feature.

The design of this implementation is inspired by `flake8`.

Effectively, what someone could do is define a package like this:

```
custom-papermill
| - setup.py
|- custom_papermill
    |- __init__.py
```

In `custom_papermill/__init__.py`, you could define a handler like this:

```python
# __init__.py

class CustomHandler(object):
    @classmethod
    def read(cls, path):
        return retrieve_custom_path(path)

    @classmethod
    def listdir(cls, path):
        raise PapermillException('listdir is not supported by CustomHandler')

    @classmethod
    def write(cls, buf, path):
        write_to_custom_path(buf, path)

    @classmethod
    def pretty_path(cls, path):
        return path
```

Then, in setup.py:

```python
# setup.py
from setuptools import setup, find_packages

setup(
    name = 'custom-papermill',
    version = '1.0.0',
    url = '',
    author = 'Author Name',
    author_email = 'author@gmail.com’,
    description = 'Description of my package',
    packages = find_packages(),
    install_requires = [],
    entry_points={
        "papermill.handlers": "customprotocol://=custom_papermill:CustomHandler"
    }
)
```

Then if a user has `custom_papermill` installed, a call like `papermill my_notebook.ipynb customprotocol://my_notebook.ipynb` will use the custom handler as defined in the custom_papermill package.

Outstanding on this:

- decide on the entry point group name (currently `papermill.handlers`, but can be changed)
- document how to create custom papermill entry points

I'm copying @zblz and @liamcoatman in on this as I've discussed this with them (we are interested in this as part of the platform SherlockML, which has a [report feature](https://docs.sherlockml.com/user-guide/reports.html).

### DIFF
--- a/papermill/engines.py
+++ b/papermill/engines.py
@@ -4,6 +4,7 @@ import datetime
 import dateutil
 
 from functools import wraps
+import entrypoints
 
 from .log import logger
 from .exceptions import PapermillException
@@ -33,6 +34,11 @@ class PapermillEngines(object):
 
     def register(self, name, engine):
         self._engines[name] = engine
+
+    def register_entry_points(self):
+        # Load handlers provided by other packages
+        for entrypoint in entrypoints.get_group_all("papermill.engine"):
+            self.register(entrypoint.name, entrypoint.load())
 
     def get_engine(self, name=None):
         """
--- a/papermill/engines.py
+++ b/papermill/engines.py
@@ -340,3 +340,4 @@ class NBConvertEngine(Engine):
 papermill_engines = PapermillEngines()
 papermill_engines.register(None, NBConvertEngine)
 papermill_engines.register('nbconvert', NBConvertEngine)
+papermill_engines.register_entry_points()
--- a/papermill/tests/test_iorw.py
+++ b/papermill/tests/test_iorw.py
@@ -22,7 +22,6 @@ from ..exceptions import PapermillException
 from ..iorw import HttpHandler, LocalHandler, ADLHandler, PapermillIO
 
 FIXTURE_PATH = os.path.join(os.path.dirname(__file__), 'fixtures')
-SAMPLE_PACKAGE = samples_dir = os.path.join(os.path.dirname(__file__), 'entrypointpackage')
 
 
 class TestPapermillIO(unittest.TestCase):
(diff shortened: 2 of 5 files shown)


papermill/tests/test_iorw.py
```diff
@@ -22,6 +22,7 @@
 from ..iorw import HttpHandler, LocalHandler, ADLHandler, PapermillIO
 
 FIXTURE_PATH = os.path.join(os.path.dirname(__file__), 'fixtures')
+SAMPLE_PACKAGE = samples_dir = os.path.join(os.path.dirname(__file__), 'entrypointpackage')
```
janfreyberg: Actually, this line is obsolete.

betatim:
I think this is a great idea! It solves the problem mentioned in #257 in a nice(r) way. It allows the base package to reduce its dependencies and users can install exactly the IO handlers they want.

I'd scope the entrypoints to allow other extensions as well as IO (for example execution engines):
```
entry_points={
        "papermill.io": "customprotocol://=custom_papermill:CustomHandler"
        "papermill.engine": "somethingsomething-here=custom_papermill_engine:CustomEngine"
    }
```

janfreyberg:
Yep, that sounds like a great idea! I'll add that in.

MSeal:
This is great contribution. Let me dig into this a bit more early this next week so I understand the capabilities here and play with an example extension, as I'm about to hop on a flight and haven't gotten to try it out yet.

janfreyberg:
> I'd scope the entrypoints to allow other extensions as well as IO (for example execution engines):
> 
> ```
> entry_points={
>         "papermill.io": "customprotocol://=custom_papermill:CustomHandler"
>         "papermill.engine": "somethingsomething-here=custom_papermill_engine:CustomEngine"
>     }
> ```

I've done this now. I think down the line it would be nice to add an abstract class for IO handlers, to allow developers to see what interface they are supposed to implement!

papermill/engines.py
```diff
@@ -34,6 +35,11 @@ def __init__(self):
     def register(self, name, engine):
         self._engines[name] = engine
 
+    def register_entry_points(self):
```
betatim: This needs calling at the end of this file I think so that entrypoints actually get added.

janfreyberg:
Great catch! I've added that in.

PR #261, opened 2018-12-03
Add support for executing notebooks in a different directory

betatim:
This adds support for executing a notebook in a different working directory. This seems nicer than having to do `(cd some/where && papermill input.ipynb out.ipynb)` or `os.chdir('some/where')` before calling `pm.execute_notebook()`.

It lets you run `papermill --cwd some/where input.ipynb out.ipynb`. `cwd` seems a bit weird as name but it is what Python's `subprocess` uses to name the argument to set the current working directory for new processes so I stuck with it.

Generally useful? If yes I'll add tests.

Created as part of work for https://www.colorado.edu/earthlab/ 🎉

### DIFF
--- a/papermill/tests/test_cli.py
+++ b/papermill/tests/test_cli.py
@@ -303,22 +303,6 @@ class TestCLI(unittest.TestCase):
             progress_bar=True,
             start_timeout=60,
             report_mode=False,
-        )
-
-    @patch(cli.__name__ + '.execute_notebook')
-    def test_prepare_only(self, execute_patch):
-        self.runner.invoke(papermill, self.default_args + ['--prepare-only'])
-        execute_patch.assert_called_with(
-            'input.ipynb',
-            'output.ipynb',
-            {},
-            engine_name=None,
-            prepare_only=True,
-            kernel_name=None,
-            log_output=False,
-            progress_bar=True,
-            start_timeout=60,
-            report_mode=False,
             cwd=None,
         )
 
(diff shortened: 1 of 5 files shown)


betatim:
Re-reading the description now I think it is not mega clear. I have notebooks in sub-directories and want to run each one in its sub-directory without having to first `cd` there (and back).
A bit like this:
```
repo/
  week1/
  week1/homework.ipynb
  week2/
  week2/homework.ipynb
```
where I want to execute the homework notebook from the `week1/` directory with `week1/` as the working directory (so relative paths in the notebook are correct).

papermill/tests/test_cli.py
```diff
@@ -293,38 +303,42 @@ def test_prepare_only(self, execute_patch):
             progress_bar=True,
             start_timeout=60,
             report_mode=False,
+            cwd=None,
         )
 
     @patch(cli.__name__ + '.execute_notebook')
-    def test_prepare_only(self, execute_patch):
-        self.runner.invoke(papermill, self.default_args + ['--prepare-only'])
```
MSeal: Did you mean to delete the prepare-only test?

papermill/tests/test_cli.py
```diff
@@ -421,22 +440,7 @@ def test_start_timeout(self, execute_patch):
             progress_bar=True,
             start_timeout=123,
             report_mode=False,
-        )
-
-    @patch(cli.__name__ + '.execute_notebook')
-    def test_report_mode(self, execute_patch):
```
MSeal: Same for meaning to remove this test?

Issue #264, opened 2018-12-06
Specifying YAML file as parameter input generates warning


nickynicolson:
Docs specify that parameters may be specified in a YAML file:

    Options:
     (snip)
      -f, --parameters_file TEXT      Path to YAML file containing parameters.

Specification of a .yaml parameters file on the command line generates the following warning:

    The specified input file (example.yaml) does not end in '.ipynb' or '.json'

The execution proceeds and parameters are read, but the warning is misleading.

MSeal:
Good catch. I'll get a fix in for that.

PR #265, opened 2018-12-07
Fixed yaml read warnings

MSeal:
Fixes https://github.com/nteract/papermill/issues/264

### DIFF
--- a/papermill/iorw.py
+++ b/papermill/iorw.py
@@ -36,32 +36,30 @@ class PapermillIO(object):
     def __init__(self):
         self.reset()
 
-    def read(self, path):
-        if not fnmatch.fnmatch(os.path.basename(path), "*.*"):
+    def read(self, path, extensions=['.ipynb', '.json']):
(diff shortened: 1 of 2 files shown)


papermill/iorw.py
```diff
@@ -227,7 +225,7 @@ def pretty_path(cls, path):
 
 def read_yaml_file(path):
     """Reads a YAML file from the location specified at 'path'."""
-    return yaml.load(papermill_io.read(path))
+    return yaml.load(papermill_io.read(path, ['.json', '.yaml', '.yml']))
```
betatim: Is this the one change that is needed to fix the bug? If no then it would be good if you could tell me how the other functions get involved. I feel like I am missing something as `read_yaml_file()` wouldn't have raised the warning that was reported.

betatim: Ok, five minutes later I now also see `papermill_io.read` in the call here ... 👀. All good.

PR #267, opened 2018-12-10
Report mode fix for Jupyterlab

tomcatling:
Jupyterlab has no 'jupyter' field in cell metadata, so the code was crashing when it tried to run 

`newcell.metadata['jupyter']['source_hidden'] = True`

in report mode. I added 

`newcell.metadata['jupyter'] = newcell.get('jupyter', {})` 

before this line to fix the issue.


MSeal:
Thanks for the fix, not sure how I missed that before.

Issue #270, opened 2018-12-14
Progress bar fails if `seaborn` is imported


bnaul:
Related to #165:  on `papermill==0.16.2`, `seaborn==0.9.0`:
```
> python -c "import seaborn as sns; import papermill as pm; pm.execute_notebook(input_path='report_mode_test.ipynb', output_path='/tmp/out.ipynb')"

HBox(children=(IntProgress(value=0, max=3), HTML(value='')))
```

Looks like the notebook environment check is not right here? It looks like `tqdm.autonotebook` does handle this case properly so maybe it would make more sense to just use that instead.

```
python -c 'import seaborn as sns; from tqdm.autonotebook import tqdm; [i for i in tqdm(range(5))]'
100%|█████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 23250.02it/s]
```

betatim:
Sounds like a good idea, especially if `tqdm` will do the tedious work of figuring out if we are in a notebook or not :)

Do you want to make a PR for this?

PR #272, opened 2018-12-17
Use tqdm.auto for progress bar

bnaul:
Closes \#270.

I'm not sure I understand what the current intended behavior for `log_output` is but I think that I preserved it...?

### DIFF
--- a/papermill/engines.py
+++ b/papermill/engines.py
@@ -1,4 +1,3 @@
-import sys
 import copy
 import datetime
 import dateutil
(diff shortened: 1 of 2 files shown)


betatim:
Did you try what happens if you execute a notebook from a notebook and set `log_output=True`? I don't remember the full history either :-/ Thinking that we might not want to edit the bar format when being executed in a notebook. Not sure if tqdm ignores the bar format for its notebook output mode?

MSeal:
Thanks for the change btw -- the bar_format code that was there predated some other changes that made it now obsolete. I just removed all that extra logic and merged it so it'll make 1.7 next week. You can test behavior by using the binder/process_highlight_dates.ipynb which tests both patterns of in and outside of notebooks. I used that file to confirm behavior with --log-output on and off.

bnaul:
Sounds great, thanks!

PR #280, opened 2019-01-09
Added argument 'injected_parameters_str' to execute_notebook function…

puma314:
It would be nice if the user calling `execute_notebook` can write arbitrary code in the injected-parameters cells in addition to having papermill inject the relevant parameters. Consider the following use-case this is useful for: 

Say `df` is a notebook parameter of type `pandas.DataFrame` that the user wants to parametrize. Since papermill only allows scalar values to be passed in when `execute_notebook` is called, the user is not able to parametrize `df` directly. However say the user wants to load `df` from a file called `numbers.csv`. Then if the injected-parameters cell contained the following code: 

```
df = "numbers.csv"
df = pandas.read_csv(df)
```

then the user could effectively parametrize `df`. However, in order to do this, the user needs to pass in 
`parameters = dict(df = "numbers.csv")` but also needs to pass in the following line of code: `df = pandas.read_csv(df)` to be put in the injected-parameters cell. If there was an argument in execute notebook such that the user could pass in the string `df = pandas.read_csv(df)` and have it appended to the end of the injected-parameters cell, then this would accomplish the task. Note that this is a better UX than having the user include a cell `df = pandas.read_csv(df)` in the notebook (without having papermill inject it) since that line of code is only applicable when `df` is being set in `injected-cell`, so then the user has to be mindful of not running this extra cell when testing locally (i.e. not running the notebook with papermill).

The new `execute_notebook` function looks as follows:

`pm.execute_notebook('path/to/input.ipynb', 'path/to/output.ipynb', parameters = dict(df = "numbers.csv"), injected_parameters_str = "df = pandas.read_csv(df)")`

In this case, the argument for the string to be put at the end of the injected-parameters cell is called `injected_parameters_str`, but very open to renaming this argument to something better. The PR itself is a very small and simple change.

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -18,6 +18,7 @@ def execute_notebook(
     input_path,
     output_path,
     parameters=None,
+    injected_parameters_str=None,
     engine_name=None,
     prepare_only=False,
     kernel_name=None,
@@ -32,6 +33,7 @@ def execute_notebook(
         input_path (str): Path to input notebook.
         output_path (str): Path to save executed notebook.
         parameters (dict): Arbitrary keyword arguments to pass to the notebook parameters.
+        injected_parameters_str (str): Arbitrary string to place at end of injected-parameters cell.
         prepare_only (bool): Flag to determine if execution should occur or not.
         kernel_name (str): Name of kernel to execute the notebook against.
         progress_bar (bool): Flag for whether or not to show the progress bar.
@@ -53,8 +55,8 @@ def execute_notebook(
     kernel_name = kernel_name or nb.metadata.kernelspec.name
 
     # Parameterize the Notebook.
-    if parameters:
-        nb = parameterize_notebook(nb, kernel_name, parameters, report_mode)
+    if parameters or injected_parameters_str:
+        nb = parameterize_notebook(nb, kernel_name, parameters, injected_parameters_str, report_mode)
 
     nb = prepare_notebook_metadata(nb, input_path, output_path, report_mode)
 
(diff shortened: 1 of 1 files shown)


MSeal:
I see the use-case here, but I am a little hesitant to immediately add this to papermill. It steps the complexity of what papermill is doing up a notch and makes the requirement on parameterizing the notebook in exactly the right way to execute harder on the caller. It also makes the planned parameter cell inspection task a little harder in the future for analyzing parameters correctly. Oh and there's a security aspect here too for certain use-cases if we expand this to the CLI where code injection, as opposed to just data injection, would be possible from clients. This isn't a blocking issue but it'd come up with some of the larger enterprise scenarios, so we may want to limit how the option can be used or make it disableable for those users.

On the otherhand it does reduce cumbersome boiler plate code needed after parameter cells. It'd definitely be convenient for wholly preparing the parameters with the correct types / complex relationships via papermill's injection. I think if we went forward I'd actually make name it `post_parameter_code` to make it clear that your injecting raw code.

@mpacer @rgbkrk @betatim @willingc Thoughts? You all may have more insight into where the boundaries should be drawn for code injection.

MSeal:
Oh also thanks for the clear description, issue, and PR -- makes having this conversation a lot easier. And welcome to nteract / papermill @puma314

rgbkrk:
The `post_parameter_code` (or `code_after_parameter_cell` if we were super verbose) seems like an _ok_ish option, since we can put it in a new cell afterward and it could be something we put on a block list in a production environment.

What would worry me about that is the impending sprawl that would likely occur -- all the template are kind of meaningless. You can now run an empty notebook, only specifying parameters and this post param cell as the entire parameterized notebook rather than insisting on breaking up code into new notebooks for re-use.

betatim:
I'm not a huge fan. It feels like it would be hard to specify useful code to be injected on the command-line/as CLI arguments. It is easier to write code and pass it as argument when using papermill as a library, so it would be helpful when building out some kind of notebooks-call-notebooks kinda workflow engine.

Overall I think I'd not add this.

Maybe the way to go is to add a hook/entrypoint/plugin that allows people to write "preprocessors" for notebooks. That way you could have a python package that implements a preprocessor that injects a cell that contains the boilerplate code you'd otherwise inject via an argument. Could be nice because you wouldn't have to repeat the code in each invocation and instead just select/activate the plugin. This is a crazy idea I thought of just now so I haven't looked around where/how we'd hook in and allow the hook to specify/consume additional command-line arguments. If I wanted to do this I'd probably investigate a bit how the pytest plugin system works (thinking of pytest-coverage which adds the `--cov=` CLI argument and collects coverage information).

MSeal:
@puma314 was the goal with this change to enable translating csv/data paths into dataframes? While it seems we don't want arbitrary code injection at the execute interface, it might make more sense to look at the translators and see about registering a custom one which is a little smarter about the json-to-code translations and auto-detects data file paths to be loaded.

Issue #227, opened 2018-10-05
Multi-environment test runner


MSeal:
When developing locally it's easy to forget to run on python 2, or the check flake for failures. We should have a way to run all the checks travis is going to run before committing or pushing.

I've used tox before successfully for this pattern. Thoughts on adding a tox definition? It doesn't block running pytest in any way.

PR #283, opened 2019-01-13
Add Tox

MSeal:
Addresses https://github.com/nteract/papermill/issues/227 so our builds and local development use the same test calls. Also helps separate manifest and linting errors from the unit testing paths so it's clearer what failed when there's such an issue.

Had some trouble with py37 seg faulting on my local machine. But I think this is a ubuntu 17.10 issue as it only bundles an alpha build of py37 from 2017 :/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,6 @@ install:
   - pip install tox
 script:
   - tox -e $TOXENV
-  - touch .coverage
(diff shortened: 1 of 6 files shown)


betatim:
Travis is happy, so it must be right. I am not a big user of tox so I'll go with what the robot says.

PR #290, opened 2019-01-23
Add None translation

k-tahiro:
This PR adds `None` translation method to `Translator` class.

When I try to pass `None` through papermill,
it's translated into `"None"` string object because `translate_escaped_str` method is called.

I think that `None` should be translated into `None` in Python notebook.


MSeal:
Thanks for the update -- this looks great. Sorry it took me a few days to get around to reviewing it.

PR #292, opened 2019-01-24
Refactor docs

willingc:
This started as a few autodoc changes but wound up being a larger refactor of the autodoc pages and the usage pages.

Overall, it breaks up the larger files into smaller files that better reflect (IMHO) the usage strengths of papermill. It also makes it easier for the reader since there is less scrolling on the page.

There's still a few things that I would like to add before NES. Links to Matt's talks, podcasts, and posts and a few more examples on binder to illustrate each workflow concept.

[Rendered docs](https://test-papermill.readthedocs.io/en/update-docs/index.html)

cc/ @MSeal @captainsafia @rgbkrk


MSeal:
The dist test failure is unrelated -- I'm going to fix that on master next week

rgbkrk:
I don't really have anything substantial to add, this looks like a good breakout of files though. Are there any content sections you'd especially like me to look at?

willingc:
@rgbkrk Nope. Just the general breakup of info. Working on adding some more examples to binder working from the doc examples and test examples but that will be separate PRs like #293

papermill/abs.py
```diff
@@ -1,10 +1,27 @@
...
 import re
 import io
+import logging
+
 from six.moves import urllib
+
 from azure.storage.blob import BlockBlobService
 
 
+logger = logging.getLogger('papermill.abs')
```
MSeal: Since we don't use this logger anywhere should this still be here?

MSeal:
Also minor flake8 issue: 
```
papermill/s3.py:397:5: E303 too many blank lines (2)
    def listdir(self, name, **kwargs):
    ^
1     E303 too many blank lines (2)
```

papermill/abs.py
willingc: We can remove this for now. We probably should better logging going forward. When we do I would `logger = logging.getLogger(__name__)`

willingc: Removed

PR #295, opened 2019-01-27
Add GCS handler

### LINKED ISSUE #213: Google Bucket File System Support
Along with the other filesystem requests, it would be amazing to be able to use papermill with Google Buckets filesystems. I'll try to contribute to this myself as well.

bnaul:
Fixes #213. This looks like all that's needed but I'm not sure what to do about testing. It seems like the S3/Azure functionality is all basically implementing stuff that's already part of `gcsfs`, hence the simplicity of this logic; I'm not sure that tests like those in `test_s3` and `test_abs` would really be doing anything...

### DIFF
--- a/papermill/iorw.py
+++ b/papermill/iorw.py
@@ -224,30 +224,26 @@ class ABSHandler(object):
 
 
 class GCSHandler(object):
-    _client = None
+    def __init__(self):
+        self._client = None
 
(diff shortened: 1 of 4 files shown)


bnaul:
@MSeal  Added a couple of simple tests and made the class -> instance variable change

PR #297, opened 2019-01-30
Cleanup version and docs

MSeal:
Should fix all the build and tox issues as well as some badge fixes.


setup.py
```diff
@@ -35,49 +47,49 @@ def read(fname):
...
-pip_message = ''
-
-try:
-    import pip
 
-    pip_version = tuple([int(x) for x in pip.__version__.split('.')[:3]])
-    pip_too_old = pip_version < (9, 0, 1)
-    if pip_too_old:
-        # pip is too old to handle IPython deps gracefully
+# Tox has a weird issue where it can't import pip from it's virtualenv when skipping normal installs
```
willingc: Hmm...didn't know that. Interesting and well commented.

.bumpversion.cfg
```diff
@@ -0,0 +1,7 @@
+[bumpversion]
```
MSeal: This needs to be in it's own file and not in setup.cfg because bumpversion clears comments from the entire file when it edits the version

MSeal:
@willingc I'm all good with swapping versioning tooling later, should be easier now with these changes.

rgbkrk:
Cool, thanks for cleaning up old decisions I made. 😄

PR #298, opened 2019-01-30
Update README.md to include supported Handlers. Minor iorw cleanup

gogasca:
- Update README.md to include supported name handlers.
- Minor cleanup in iorw.py to conform to previous style during instantiation of PapermillIO.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ Papermill supports the following name handlers for input and output paths during
 
  * Azure: [Azure DataLake Store](https://docs.microsoft.com/en-us/azure/data-lake-store/data-lake-store-overview), [Azure Blob Store](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blobs-overview) `adl://, abs://`
     
- * Google Cloud: [Google Cloud Storage](https://cloud.google.com/storage/) `gcs://, gs://`
+ * Google Cloud: [Google Cloud Storage](https://cloud.google.com/storage/) `gs://`
 
 Python In-notebook Bindings
 ---------------------------
--- a/README.md
+++ b/README.md
@@ -140,6 +140,20 @@ linear_function:
     intercept: 1.0"
 ```
 
+#### Supported Name Handlers
+
+Papermill supports the following name handlers for input and output paths during execution: 
+
+ * Local file system: `local`
+
+ * HTTP, HTTPS protocol:  `http://, https://`
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -140,6 +140,20 @@ linear_function:
     intercept: 1.0"
 ```
 
+#### Supported Name Handlers
```
MSeal: Thanks -- a welcome addition. If you felt motivated adding it to the docs/ would be helpful too (not necessary for merge)

papermill/iorw.py
```diff
@@ -255,9 +255,8 @@ def pretty_path(self, path):
 papermill_io.register("abs://", ABSHandler)
 papermill_io.register("http://", HttpHandler)
 papermill_io.register("https://", HttpHandler)
-gcs_handler = GCSHandler()
-papermill_io.register("gcs://", gcs_handler)
-papermill_io.register("gs://", gcs_handler)
+papermill_io.register("gcs://", GCSHandler())
```
MSeal: Do you want each end-point type (gs vs gcs) to share the cached client or is having two separate instances ok by default?

gogasca: I just checked with Cloud Storage team and @frankyn and we only support gs:// paths.

```
~ $ gsutil ls -al gs://dpe-sandbox/notebook.ipynb
     41946  2019-01-26T09:44:36Z  gs://dpe-sandbox/notebook.ipynb#1548495876399041  metageneration=1
TOTAL: 1 objects, 41946 bytes (40.96 KiB)
~ $ gsutil ls -al gcs://dpe-sandbox/notebook.ipynb
InvalidUrlError: Unrecognized scheme "gcs".
```
Documentation [here](https://cloud.google.com/storage/docs/listing-objects)

Would be interested in asking @bnaul the reason for adding gcs://, otherwise we can just keep gs://

README.md
gogasca: Definitely will do.

MSeal:
Thanks. Going to merge for now. If you get the docs/ update we can merge that later.

bnaul:
No problem from me, `gcsfs` supports both so I copied that: https://gcsfs.readthedocs.io/en/latest/#connection-with-dask-and-zarr. I've never actually seen `gcs://` used personally

gogasca:
@bnaul Thank you, thank you very much for this PR, this is very helpful for us.

Issue #266, opened 2018-12-10
Documenting entry points.


janfreyberg:
Since entry points are in the [latest release](https://github.com/nteract/papermill/commit/29ccb408d5865f950f019b5aa3f5a2640ca23aeb) it would be nice to document how to develop new entry points so third-party developers can do so.

I will take a stab at this, but I wanted to open an issue to discuss the format. At the moment, there are two main sections in the docs: *Installation* and *Usage*. My proposal would be to add a third section titled e.g. *Developing papermill* that has general development advice for the library overall, and specific sub-sections for developing an engine and an IO handler.

Thoughts?

betatim:
Sounds like a good plan.

"Developing" is totally fine and should be our baseline. Pondering if by using a slightly different word we can increase the number of people who think changing/tweaking papermill is something they can and should do? Some random words that might be suitable: extending, developing, contributing, modifying, changing ... papermill.

janfreyberg:
I like *extending* or *customising*, that's nice and accessable!

MSeal:
Yes please! That'd be a great contribution.

willingc:
Perhaps the outline should be similar to:
- Extending papermill
    - High level description and visual about engines and io
    - New engines
         - How to develop a new engine
         - How to install and use
         - A runnable example or template
     - New IO handler
         - How to develop
         - How to install and use
         - Example or template

janfreyberg:
I like that! I will try and finally make a start on this today.

willingc:
Thanks @janfreyberg. Just an FYI that I have an open PR #292 for a docs refactor. The docs for Extending papermill would be great after the `usage-cli` section. Please let me know if you need anything too.

PR #299, opened 2019-02-04
Added isolated depedency installation options

MSeal:
Implements https://github.com/nteract/papermill/issues/257

My only worry with merging this is around existing users who have `papermill` as a dependency setup assuming s3 or azure bindings are available. This might mean we should wait for 1.0 release? Gcs users did want their dependencies isolated from standard installations before releasing that plugin though.


betatim:
Who are Gcs users?

betatim:
I'd keep S3 and Azure bundled/installed by default for two (?) releases to give people a chance to update their dependencies.

Ideal would be to emit a `DeprecationWarning` to let people know the change is coming but I am not quite sure if we can detect if they explicitly used `pip install papermill[s3]` or just `pip install papermill`.

betatim:
In the last release IO and execution engine "plugins" were enabled by using entrypoints: https://github.com/nteract/papermill/issues/266. I think that is a good way of allowing people to extend papermill. 

Not sure I understand what you mean with "I'd need to create a new setup.py, etc. which I think is a bit heavier-weight and might affect adoption of building custom extensions". I think if you want `pip install my-awesome-engine` to work you need a `setup.py`. Maybe we should provide a template for the `setup.py` as part of the docs in #266 to make it easier for those not familliar with Python packaging to do the right thing.

betatim:
To make travis happy we need to install the Azure related packages in the test environment. I think that is better than removing the tests.

mckev-amazon:
Good point - after reading the document I referenced, I realize that the burden is on the plugin developer only to configure setup.py.  In my hastiness I incorrectly assumed that this was an end-user responsibility - indeed it's not.  In that regard, providing documentation to the plugin developer is certainly reasonable.  As you suggest, a setup.py template would be helpful.

I suppose based on that I'm still not entirely clear on, if we had to pick one approach, which is the most suitable.  I couldn't find anything that suggests that one is more canonical over the other.  If we want to optimize for adoption of new engine implementations, the first two approaches on the surface could be considered more approachable, on the other hand, entrypoints is more explicit and provides more flexibility as it allows one package to expose multiple entrypoints. 

All that said, as I'm typing this I'm realizing I'm probably splitting hairs at this point :-)

Issue #306, opened 2019-02-08
On Windows only, papermill can't be imported unless the env variable APPDATA is set


mgasner:
This is because papermill.iorw, by default, imports the Azure Datalake facilities:

```
...
    import papermill as pm
...\lib\site-packages\papermill\__init__.py:8: in <module>
    from .api import display, record, read_notebook, read_notebooks
...\lib\site-packages\papermill\api.py:17: in <module>
    from .iorw import load_notebook_node, list_notebook_files
...\lib\site-packages\papermill\iorw.py:20: in <module>
    from .adl import ADL
...\lib\site-packages\papermill\adl.py:2: in <module>
    from azure.datalake.store import core, lib
...\lib\site-packages\azure\datalake\store\__init__.py:11: in <module>
    from .core import AzureDLFileSystem
...\lib\site-packages\azure\datalake\store\core.py:29: in <module>
    from .utils import ensure_writable, read_block
...\lib\site-packages\azure\datalake\store\utils.py:21: in <module>
    datadir = os.path.join(os.environ['APPDATA'], 'azure-datalake-store')
...\lib\os.py:425: in __getitem__
    return self.data[key.upper()]
E   KeyError: 'APPDATA'
```

The issue is this switch in https://github.com/Azure/azure-data-lake-store-python/blob/master/azure/datalake/store/utils.py#L20.

Happy to open a pull request which wraps the ADL import in a try/except.

MSeal:
If you look at 0.18.0 release that went out today papermill no longer installs azure by default, so it would only hit this error if you did install with `papermill[azure]`.

That being said I would appreciate an extension to the import failures in iorw.py to capture KeyError against `APPDATA` and something akin to the `missing_dependency_generator`.

mgasner:
Thanks so much. I have #307 open for you.

PR #307, opened 2019-02-09
Catch missing APPDATA key

mgasner:
Addresses #306


MSeal:
Perfect -- thank you for the improvement.

PR #313, opened 2019-02-15
Handle 429 Errors

gogasca:
Add retry library to be able to handle writes to GCS.
Currently papermill emits a save after each cell executes.

### DIFF
--- a/papermill/iorw.py
+++ b/papermill/iorw.py
@@ -11,6 +11,7 @@ import nbformat
 import requests
 import warnings
 import entrypoints
+import retry
 
 from contextlib import contextmanager
 
@@ -46,6 +47,11 @@ try:
     from gcsfs import GCSFileSystem
 except ImportError:
     GCSFileSystem = missing_dependency_generator("gcsfs", "gcs")
+
+try:
+    from gcsfs.utils import HtmlError as RateLimitException
+except ImportError:
+    RateLimitException = Exception
 
 try:
     FileNotFoundError
@@ -255,6 +261,7 @@ class GCSHandler(object):
     def listdir(self, path):
         return self._get_client().ls(path)
 
+    @retry.retry(RateLimitException, delay=1, backoff=2, max_delay=4)
     def write(self, buf, path):
         with self._get_client().open(path, 'w') as f:
             return f.write(buf)
--- a/papermill/iorw.py
+++ b/papermill/iorw.py
@@ -261,7 +261,7 @@ class GCSHandler(object):
     def listdir(self, path):
         return self._get_client().ls(path)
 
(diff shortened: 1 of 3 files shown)


papermill/iorw.py
```diff
@@ -255,6 +261,7 @@ def read(self, path):
     def listdir(self, path):
         return self._get_client().ls(path)
 
+    @retry.retry(RateLimitException, delay=1, backoff=2, max_delay=4)
```
MSeal: Could I convince you to add a test which raises RateLimitException to prove it retries?

MSeal: Also do we need to handle non 429 errors differently? Probably not blocking but ideally we'd capture only the 429 exceptions. Not blocking a merge but thought I'd check.

gogasca: @MSeal Will add test.
Agree we probably need to handle other errors differently (No write access/invalid path, etc)

requirements.txt
```diff
@@ -13,4 +13,4 @@ jupyter_client
 pandas
 requests
 entrypoints
-
+retry
```
betatim: This is a neat library that I didn't know existed!

papermill/iorw.py
gogasca: 1. I added a test. (May need some review)
2. I created a new PR on gcsfs. https://github.com/dask/gcsfs/pull/131 this will return the specific Exception on 429 error. We can probably wait and I can make the required changes, let me know how you want to proceed.
Thanks

papermill/tests/test_gcs.py
```diff
@@ -42,7 +47,20 @@ def test_gcs_write(mocker):
...
+
 
+def test_gcs_write_retry(mocker):
+    mocker.patch('papermill.iorw.GCSFileSystem', MockGCSFileSystem)
+    counter = 0
+    try:
+        gcs_handler = GCSHandler()
+        client = gcs_handler._get_client()
+        gcs_handler.write('', 'gs://bucket/test.ipynb')
+    except RetryableError:
```
MSeal: I don't think this actually tests the retry code path. RetryableError isn't being caught by the handler and automatically rerun because it's not a RateLimitException.

I think there should be one test which tests that retries can be exhausted by returning RateLimitException indefinitely and count the number of write attempts within the MockFCS object. Then another test in which MockFCS raises RateLimitException once and then succeeds.

MSeal: Also you can test that exceptions occurred without a try/except by doing:
```
with self.assertRaises(MyExceptionName):
  code_which_raises_the_exception()
```

gogasca: Rewrite the tests. PTAL Thanks

PR #315, opened 2019-02-17
Added binder check to tox tests

MSeal:
Now you'll know the binder link will work before merging :)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -26,7 +26,7 @@ commands =
 
 # Binder
 [testenv:binder]
-description = execute all notebooks in binder to ensure they're runnable
+description = ensure /binder/*ipynb are runnable
 changedir = binder
 deps =
     -r binder/requirements.txt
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -24,6 +24,14 @@ commands =
     sphinx-build -d "{toxworkdir}/docs_doctree" docs "{toxworkdir}/docs_out" --color -W -bhtml {posargs}
     python -c 'import pathlib; print("documentation available under file://\{0\}".format(pathlib.Path(r"{toxworkdir}") / "docs_out" / "index.html"))'
 
+# Binder
+[testenv:binder]
+description = execute all notebooks in binder to ensure they're runnable
```
betatim: Should we change this to something like "execute all notebooks in the environment used on Binder'? When I read this I thought the test would be making calls to mybinder.org and we confused that it wasn't.

betatim:
One small naming comment, otherwise 🚢 it.

willingc:
Nice @MSeal

PR #319, opened 2019-02-21
Handle Python 3.5 version of Python not handling bytes object in json…

gogasca:
… loads

- Add conversion of bytes object to str via decode('utf-8')
- Add test case
- Add GCS notebook

### DIFF
--- a/papermill/tests/test_iorw.py
+++ b/papermill/tests/test_iorw.py
@@ -64,7 +64,6 @@ class TestPapermillIO(unittest.TestCase):
             self.ver = ver
 
         def read(self, path):
-
             local_dir = os.path.dirname(os.path.abspath(__file__))
             with open(os.path.join(local_dir, path)) as f:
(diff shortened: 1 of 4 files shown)


papermill/tests/test_iorw.py
```diff
@@ -48,6 +48,10 @@ def __init__(self, ver):
             self.ver = ver
 
         def read(self, path):
+            if path == 'fake/path_bytes':
```
MSeal: Let's not make this so complicated in the test (plus this isn't testing that the code in iorw module actually does the decode you repeat here). Instead make a FakeByteHandler class which reads and returns the notebook you attached above. Then make a special test for reading from that handler and proving it does not raise an exception. This will test your code path cleanly.

gogasca: Thanks for the review, submitted new test, PTAL.

PR #322, opened 2019-02-26
Add extending papermill documentation

janfreyberg:
This adds documentation for how to write custom entry points for I/O handlers or engines, as well as a short page on how to develop papermill itself, pointing to the github guides.

This addresses #266. I couldn't think of a simple example implementation of a custom engine, so I haven't provided an example for that... any ideas welcome!


docs/extending-entry-points.rst
```diff
@@ -0,0 +1,224 @@
...
+Virtually the first thing that happens when papermill is used is that the input
+notebook is read in. This is managed by I/O handlers, which allow papermill to
+access not just the local filesystem, but also remote services such as Amazon
+S3. The same goes for writing the executed notebook to a file system: I/O
+handlers allow papermill to write files to S3 or otherwise.
+
+Creating a new handler
+~~~~~~~~~~~~~~~~~~~~~~
+
+Writing your own I/O handler simply requires writing a class that has four
```
betatim:
```suggestion
Writing your own I/O handler requires writing a class that has four
```
I'd avoid words like "simply" or "just do X" on the one hand it makes it sounds like it should be simple (and if you don't find it simple you are not smart enough) and on the other hand it annoys the anti-[weasel word](https://en.wikipedia.org/wiki/Weasel_word) people.

docs/extending-entry-points.rst
```diff
@@ -0,0 +1,224 @@
...
+
+``nb_man`` is a |nbformat.NotebookNode|_, and ``kernel_name`` is a string. Your
+custom class then needs to implement the execution of the notebook. For example,
+you could insert code that executes the notebook remotely on a server, or
+executes the notebook many times to simulate different conditions.
+
+Ensuring your engine is found by papermill
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Much like I/O handlers, engines can be specified as entry-points, under the
```
betatim: I'd write this assuming the reader arrives here without having first read the I/O handler section. Another pet peeve of mine is docs that assume you've read all of it when you are trying to use it as a reference.

betatim:
This is excellent! I left a few comments with things that are more a question of taste/style than "must haves" or things I'd insist on.

janfreyberg:
Good points! I'll update those.

I think the tests failed because of a change in `gcsfs`:

https://github.com/dask/gcsfs/commit/58a8990c733519a9055136c1f68ba4b1ab4b24e7

MSeal:
I'll get papermill fixed for the gcsfs change independently

MSeal:
Got a bunch of things I am juggling -- will try to get a thorough read through on this while traveling, but at first glance this is great.

An silly example for a custom engine might be one which only syncs on notebook completion instead of cell and notebook completion. Or one which adds extra logging for each message received from the kernel, or metadata change, etc.

MSeal:
Fixed the tests fyi

docs/extending-entry-points.rst
janfreyberg: Agreed! I've updated this here and elsewhere.

docs/extending-entry-points.rst
janfreyberg: Yep - I've amended this.

janfreyberg:
Good ideas! I've gone with an example that simply adds some of the metadata as output to each code cell, to make it a bit more visual. I think this would be ready for review again now!

betatim:
Merged! This is a great PR and I enjoyed reviewing it, especially because of the way you went about things 👍.

PR #325, opened 2019-03-01
adding to_pdf export function

sinemetu1:
Is there any interest for adding functionality like this? I spent some time trying to figure out whether I could pass parameters or some kwargs to papermill to get it to export to pdf but didn't find anything so I put together this function instead.

### DIFF
--- a/papermill/__init__.py
+++ b/papermill/__init__.py
@@ -2,6 +2,6 @@ from __future__ import absolute_import, division, print_function
 
 from .version import version as __version__
 
-from .api import display, record, read_notebook, read_notebooks
+from .api import display, record, read_notebook, read_notebooks, to_pdf
 from .exceptions import PapermillException, PapermillExecutionError
 from .execute import execute_notebook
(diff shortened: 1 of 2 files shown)


MSeal:
Ahh sorry to reject this, but we definitely want to keep papermill as notebook-to-notebook execution. In the past we've recommended people combine --prepare-only to parameterize notebooks with direct nbconvert calls to translate to pdf/html/etc...

sinemetu1:
Hm, isn't `nbconvert` being used under the covers when `execute_notebook` is called? It would make sense to add the ability to pass full parameters to `nbconvert` in this case if an api method like `to_pdf` isn't an option otherwise there are these workarounds.

MSeal:
The problem is two-fold. We want papermill to stay as simple as we can. It's lightweight and easy to reason about today. Nbconvert is the Swiss army knife for notebook conversions and there's not a need to replicate it's functionality in entirety. The second problem is that the engine to execute notebooks is abstract, even though there isn't other engines in the open source version of the software yet. We'd need to either ensure other engines can make the same calls or have an interface that may fail by design under certain conditions. Overall it's a design decision we make last year about the direction of the library. Hopefully that makes some sense. And I do feel bad closing your PR. Generally we don't directly reject changes

PR #333, opened 2019-03-21
Log a warning if the input notebook doesn't have a parameters cell

### LINKED ISSUE #245: Error on executing a notebook without parameter tag
I sometimes (more often than I want to admit) run `papermill a-nb-withot-parameter-cell.ipynb output.ipynb -p alpha 0.1` mostly because I edited the notebook in nteract and forget to press save or get the paths wrong.

Should it be an error/warning to execute a notebook that has no parameters and specify parameters on the command-line?

Madhu94:
Should close #245, let me know if somethine else was needed


Madhu94:
Fixed the test

Issue #300, opened 2019-02-04
Add path interpolation using parameters and special keys


MSeal:
papermill highlight_dates.ipynb ./highlight_dates_run_cli_{start_date}_{stop_date}_{%guid}.ipynb -p start_date 2011-01-01 -p stop_date 2014-02-02

Use string / int parameters and a few built-in keywords to generating notebook paths,

VincentBLortie:
I like that the example looks like python string formatting: most users will be familiar with the syntax. Perhaps there's another way than % to signify that a parameter is a built-in? (afaik, % isn't something you can use in a python format string). How about a prefix like "pm_" for values automatically provided by papermill?

Actually, from an implementation point of view, do you see any danger in using the user's input as a format string and calling .format directly?

MSeal:
I think format syntax would be the clearest and easiest. We may want to raise an error if a format string isn't filled in would be the only thing to keep in mind.

PR #343, opened 2019-04-15
Add path interpolation using parameters and special keys

VincentBLortie:
Work in progress for nteract/papermill#300

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -2,26 +2,15 @@
 
 from __future__ import unicode_literals, print_function
 
-import six
 import copy
 import nbformat
 
 from .log import logger
-from .exceptions import PapermillExecutionError, PapermillMissingParameterException
-from .iorw import (
-    load_notebook_node,
-    write_ipynb,
-    read_yaml_file,
-    get_pretty_path,
-    local_file_io_cwd,
-)
-from .translators import translate_parameters
+from .exceptions import PapermillExecutionError
+from .iorw import load_notebook_node, write_ipynb, get_pretty_path, local_file_io_cwd
 from .engines import papermill_engines
 from .utils import chdir
-from string import Formatter as StringFormatter
-from collections import OrderedDict
-from uuid import uuid4
-from datetime import datetime
+from .parameterize import parameterize_notebook, parameterize_path, add_builtin_parameters
 
 
 def execute_notebook(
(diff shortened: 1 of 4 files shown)


papermill/execute.py
```diff
@@ -136,6 +141,33 @@ def prepare_notebook_metadata(nb, input_path, output_path, report_mode=False):
...
+    except KeyError:
+        # OrderedDict removes duplicates while maintaining order
+        missing_parameters = OrderedDict.fromkeys(
+            [
+                required_parameter
+                for (_, required_parameter, _, _) in StringFormatter().parse(path)
+                if (required_parameter is not None) and (required_parameter not in parameters)
+            ]
+        ).keys()
+        raise Exception("Missing parameters: {}".format(", ".join(missing_parameters)))
```
MSeal: Maybe `PapermillException` or extend that to `PapermillMissingParameterException`

papermill/tests/test_parameterize.py
```diff
@@ -72,3 +72,52 @@ def test_repeated_run_no_parameters_tag(self):
...
+
+    def test_path_with_numeric_parameter(self):
+        self.assertEqual(parameterize_path("foo/bar/{baz}", {"baz": 42}), "foo/bar/42")
+
+    def test_path_with_numeric_format_string(self):
+        self.assertEqual(parameterize_path("foo/bar/{baz:03d}", {"baz": 42}), "foo/bar/042")
+
+    def test_path_with_float_format_string(self):
+        self.assertEqual(parameterize_path("foo/bar/{baz:.03f}", {"baz": 0.3}), "foo/bar/0.300")
+
```
MSeal: We should maybe test list and dict inputs, since those are common parameter types.

papermill/execute.py
VincentBLortie: Good idea, will do

VincentBLortie:
What's left: built-in keywords were mentioned in #300. I am not quite sure how they should be handled yet. For example:

- How they should be referenced: Can it just be a key with a special prefix (e.g. pm_guid, pm_current_date) that is automatically added to parameters?
- Which ones to include? You had given the example of a GUID, I was thinking perhaps the current date could also be useful (local time? gmt? perhaps both) Perhaps we could start with guid, current_date and current_date_gmt and add more if we think of useful ones to have.
- I imagine that if the guid built-in is used in more than one place, such as, for example, both the input and output path, then we want to use the same generated guid in both locations

MSeal:
> What's left: built-in keywords were mentioned in #300. I am not quite sure how they should be handled yet. For example:
> 
> * How they should be referenced: Can it just be a key with a special prefix (e.g. pm_guid, pm_current_date) that is automatically added to parameters?

Maybe prefix with `run` (`pm_run_guid`) to indicate it's no unique for each reference? And make sure it is a v4 guid we inject into `parameterize_path`'s copy  of parameters? I'm not sure if we should add it to `parameters` in general as this would change the empty parameter use-case.

> * Which ones to include? You had given the example of a GUID, I was thinking perhaps the current date could also be useful (local time? gmt? perhaps both) Perhaps we could start with guid, current_date and current_date_gmt and add more if we think of useful ones to have.

That sounds reasonable. Should we keep them namespaces to a dict we inject? So you'd reference by `pm[current_date_local]` and `pm[run_guid]` so it's clearer that papermill is providing these values?

> * I imagine that if the guid built-in is used in more than one place, such as, for example, both the input and output path, then we want to use the same generated guid in both locations

Yes. I agree.

VincentBLortie:
Just found a bug with the missing parameters exception message when there are dictionary parameters. I don't know how simple it will be to report which keys are missing.

MSeal:
LGTM
Simply raising error with the message as you did seems fine to me here.
Any further concerns or would you like a merge?

VincentBLortie:
It's ready to be merged as far as I'm concerned 👍

PR #351, opened 2019-05-01
pass through notebook engine arguments

vizhur:
Allow user to pass notebook engine arguments via kwargs on pm.execute_notebook()
Right now there is no way (*I could find) to pass args like allow_errors and timeout

### DIFF
--- a/papermill/engines.py
+++ b/papermill/engines.py
@@ -358,7 +358,7 @@ class NBConvertEngine(Engine):
         # Rearrange arguments. Handle potential duplicates from **kwargs.
         # Prioritize values passed to the function or set in the module.
         preprocessor_handled_args = ['timeout', 'start_timeout', 'kernel_name', 'log']
-        safe_kwargs = {key: kwargs[key] for key in kwargs if key not in preprocessor_handled_args }
+        safe_kwargs = {key: kwargs[key] for key in kwargs if key not in preprocessor_handled_args}
         timeout = execution_timeout if execution_timeout else kwargs.get('timeout')
         preprocessor = PapermillExecutePreprocessor(
             timeout=timeout,
--- a/papermill/engines.py
+++ b/papermill/engines.py
@@ -368,7 +368,7 @@ class NBConvertEngine(Engine):
             **safe_kwargs
         )
         preprocessor.log_output = log_output
-        preprocessor.preprocess(nb_man, kwargs)
+        preprocessor.preprocess(nb_man, safe_kwargs)
 
 
 # Instantiate a PapermillEngines instance, register Handlers and entrypoints
--- a/papermill/engines.py
+++ b/papermill/engines.py
@@ -357,7 +357,7 @@ class NBConvertEngine(Engine):
         """
         # Rearrange arguments. Handle potential duplicates from **kwargs.
         # Prioritize values passed to the function or set in the module.
-        preprocessor_handled_args = ['timeout', 'start_timeout', 'kernel_name', 'log']
+        preprocessor_handled_args = ['timeout', 'startup_timeout', 'kernel_name', 'log']
         safe_kwargs = {key: kwargs[key] for key in kwargs if key not in preprocessor_handled_args}
         timeout = execution_timeout if execution_timeout else kwargs.get('timeout')
         preprocessor = PapermillExecutePreprocessor(
--- a/papermill/engines.py
+++ b/papermill/engines.py
@@ -355,11 +355,17 @@ class NBConvertEngine(Engine):
         by `nbconvert`, and it is somewhat misleading here. The preprocesser
         represents a notebook processor, not a preparation object.
         """
+        # Rearrange arguments. Handle potential duplicates from **kwargs.
+        # Prioritize values passed to the function or set in the module.
+        preprocessor_handled_args = ['timeout', 'start_timeout', 'kernel_name', 'log']
+        safe_kwargs = {key: kwargs[key] for key in kwargs if key not in preprocessor_handled_args }
+        timeout = execution_timeout if execution_timeout else kwargs.get('timeout')
         preprocessor = PapermillExecutePreprocessor(
-            timeout=execution_timeout,
+            timeout=timeout,
             startup_timeout=start_timeout,
             kernel_name=kernel_name,
             log=logger,
+            **safe_kwargs
         )
         preprocessor.log_output = log_output
         preprocessor.preprocess(nb_man, kwargs)
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -25,6 +25,7 @@ def execute_notebook(
     start_timeout=60,
     report_mode=False,
     cwd=None,
+    **kwargs
 ):
     """Executes a single notebook locally.
 
@@ -52,6 +53,8 @@ def execute_notebook(
         Flag for whether or not to hide input.
     cwd : str, optional
         Working directory to use when executing the notebook
+    **kwargs
+        Arbitrary keyword arguments to pass to the notebook engine
 
     Returns
     -------
@@ -91,6 +94,7 @@ def execute_notebook(
                     progress_bar=progress_bar,
                     log_output=log_output,
                     start_timeout=start_timeout,
+                    **kwargs
                 )
 
             # Check for errors first (it saves on error before raising)
--- a/papermill/tests/test_engines.py
+++ b/papermill/tests/test_engines.py
@@ -409,9 +409,14 @@ class TestNBConvertEngine(unittest.TestCase):
                 self.assertNotEqual(self.nb, nb)
 
                 pep_mock.assert_called_once()
-                pep_mock.assert_called_once_with(
-                    timeout=1000, startup_timeout=30, kernel_name='python', log=logger
-                )
+
+                args, kwargs = pep_mock.call_args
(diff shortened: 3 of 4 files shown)


MSeal:
Moving slack conversation on topic here for visibility for other folks.

I don't see any harm in adding this extension of pushing kwargs down as we already pass them to the engine so users can pass arguments they upstream construct without needing to monkey patch out classes.

Only thing I'd say is we should add a test for the safe_keyword logic. If we want to reuse that pattern in other kwarg passers we can make a function wrapper in utils.py down the road (not needed for this PR). 

> It might against "by design" so I want to confirm :slightly_smiling_face:. Basically, I want to set cell execution timeout and allow_errors props nbconvert provides. The way I've done this in my PR is passing through kwargs. That works 'as expected' with nb engine. 2 test cases that expect unkowned args to be dropped are failing.
> My question is, are you ok allowing user to pass arbitrary parameters to the engine? Other option could be to add those args to the execute_notebook()
> I can workaround this by implementing own engine to set the props I want, but that's not a generic solution
> 
> from papermill.engines import NBConvertEngine, papermill_engines
> from papermill.preprocess import PapermillExecutePreprocessor
> from papermill.log import logger
> class NBConvertEngine2(NBConvertEngine):

MSeal:
Fixing up the tests to expect those kwargs would be fine as well imo.

betatim:
Are we trying to filter out explicitly defined kwargs defined in `execute_managed_notebook` or `PapermillExecutePreprocessor`? For example `kernel_name` is an argument to `execute_managed_notebook` but not a keyword argument. It is however a keyword argument for `PapermillExecutePreprocessor`. In conclusion: I think we need to explain a bit better what is meant to happen here.

Also consider this example of how kwargs and args etc work:
```ipython
In [68]: def execute(kernel_name, **kwargs):
    ...:     preprocess(kernel_name=kernel_name, **kwargs)
    ...:

In [69]: def preprocess(kernel_name='kernel_name', **kwargs):
    ...:     print(kernel_name, kwargs)
    ...:

In [70]: execute('python')
python {}

In [71]: execute(kernel_name='python')
python {}

In [72]: execute(kernel_name='python', foo='bar')
python {'foo': 'bar'}
```

`execute` takes a kernel name parameter (not defined as a keyword arg) and call preprocess where it is. We don't have to do any merging/processing of the args, Python "just does the right thing". If you do manage to somehow sneak a `kernel_name` key into the `kwargs` dictionary using in the call of `preprocess` Python will raise an exception saying that the argument was passed twice.

vizhur:
@betatim 
Hi Tim,

I'm trying to pass down to the engine arguments specified by a user. args and kwargs work correctly no matter how you pass them as long as you consistent with the names. 
When it comes to PapermillExecutePreprocessor() execution_timeout is redefined as timeout and .logger set as log (because nbconvert aparently has its own names:)). That's where we need to handle potential collision.
So basically only those two should be handled: 
- ignore any logger from kwargs 
- prioritize execution_timeout over timeout if both set

betatim:
Why not like this then:

```python
  # prioritize execution_timeout if it is set, otherwise check for `timeout` in kwargs
  if execution_timeout is None:
    execution_timeout = kwargs.get("timeout")
  preprocessor = PapermillExecutePreprocessor(
      timeout=execution_timeout,
      startup_timeout=start_timeout,
      kernel_name=kernel_name,
      log=logger,
      **kwargs
  )
```

`logger` isn't an argument of `execute_managed_notebook`, it is a "global". If someone calls `execute_managed_notebook(..., log=foo)` then there will be an exception because `PapermillExecutePreprocessor` will be passed two arguments called `log`. I think using the built into Python mechanism for this and raising an exception is better than hiding the error by swallowing the argument (`log`) if it is set. 

If you call `execute_managed_notebook(..., logger=foo)` then it will (probably) get ignored because it isn't an argument that `PapermillExecutePreprocessor` knows what to do with (or maybe it will raise an exception).

vizhur:
@betatim 
Your code would fail exactly because of the issue you concern about - argument passed twice. 
I tried to keep current behavior, ie not breaking any existent scenarios and allow user to pass parameters (read set attributes) in ExecutePreprocessor.
If user call execute_managed_notebook(, log=foo) he/she will indeed get an exception without my changes as I clean duplicates in kwars before passing them down to PapermillExecutePreprocessor.
If you call execute_managed_notebook(..., logger=foo) nothing will happen it will just set the attribute on ExecutePreprocessor class
Please note, that I made kwargs changes on pm.execute_notebook only, the rest of the code was already passing kwargs, all I made there is cleaned kwargs to avoid runtime error and preserve current behavior

betatim:
> If user call execute_managed_notebook(, log=foo) he/she will indeed get an exception 

I agree. I also think this is what should happen. In your code you remove `log` from the `kwargs` passed to `execute_managed_notebook`. So if a user calls `execute_managed_notebook(..., log=foo)` there will be no exception or indication that the thing the user was trying to do (pass an argument called `log` through to the engine) failed.

Another thing I don't understand is why you [remove `kernel_name` from the `kwargs` passed to `execute_managed_notebook`](https://github.com/nteract/papermill/pull/351/files#diff-04f9b6a07a4c30ba663a32562bbfb4b9R361). The `kwargs` dictionary will not contain a key with that name because there is an argument called `kernel_name` defined in `execute_managed_notebook()`'s signature already. This means you don't need to remove it or the fact that you do attempt to remove it means that there is something I don't understand about what your code is trying to do.
```ipython
In [1]: def f(a, **kwargs):
   ...:     print(a, kwargs)
   ...:

In [2]:

In [2]: f(a='23')
23 {}

In [3]: f(a='23', b='b')
23 {'b': 'b'}
```

I think my main problem is that I don't understand how the goal you want to achieve and the code that is in this PR fit together. This usually means we need more comments or different code layout because if I can't understand what it is meant to do then that isn't a good sign as others will probably face the same problem. I like the idea of passing arguments down to the engine so let's try and get to a shared understanding of what the goal is and then how to implement it.

vizhur:
execute_managed_notebook(..., log=foo) now gives us consistent behavior, I think. The same way you pass smth=foo or nonesense=bar - it will be silently ignored if not handled in ExecutePreprocessor

kernel_name, it will never be in the kwargs in execute_managed_notebook as it will be recognized as an argument.

>>> def foo(a,**kwargs):
...     print(a)
...     print(kwargs)
...
>>> foo(1, a=3)
TypeError: foo() got multiple values for argument 'a'

vizhur:
@betatim 
The main goal is to allow user to pass engine args from pm.execute_notebook(). My bigest interest is in allow_errors and timeout. 
Current infrastructure already passes keyworded arguments, except the "top level" function pm.execute_notebook() that seems like the main user's entry point.

The first part of the changes is to add *kwargs to pm.execute_notebook and pass it down to pe.execute_notebook_with_engine(). That accomplishes the goal and makes me happy :)

The second part is not related to the end goal. It just improves user experience and makes handlind of keyworded arguments consistent. The way ExecutePreprocessor, and all derived classes constructed, sets/overwrites passed arguments to a constructor as attributes[see ln1060 here https://github.com/ipython/traitlets/blob/master/traitlets/traitlets.py]

So by design you can pass arbitrary parameters there. As PM has a bit different from NbConvert names for some of the parameters in NBConvertEngine::execute_managed_notebook they mapped back to the NBConvert-known arguments
startup_timeout<-start_timeout
timeout<-execution_timeout
Also, as PM controls notebook execution it sets own logger as log=papermill.log.logger and that makes total sense.

Mapping parameters leads to runtime error "got multiple values for argument" when user specify one of startup_timeout, timeout or log parameters. To handle that I've added the following logic

log - ignored, as described above. User have other options to pass their logs if needed
startup_timeout - use start_timeout value that is user visible on all the levels including pm.execute_notebook(). If user pass startup_timeout parameter it will be ignored as any other unknown option
timeout - takes default None of none of timeout or execution_timeout specified by a user otherwise user's defined option will be used and execution_timeout will be prioritized over timeout

MSeal:
I'd make the argument we should make a strong opinion for controlling keyword changes with regard to passing remaining arguments downstream. But I don't think that should perhaps be outside this PR, as @vizhur already pointed out we pass naked kwargs around a lot already for convenience.

I want to keep it flexible without too many sharp edges for users to make mistakes by accident. For example, one strategy I used in another setting was to make a context wrapper that logs unused arguments from kwargs.

@betatim should we start an issue thread to discuss kwarg conventions and solve that with documentation / maybe code changes to accommodate that opinion? I'm ok with the code as is without that stronger opinion established (we can update this code later if we decide to change implementations a bit).

papermill/engines.py
```diff
@@ -355,11 +355,17 @@ def execute_managed_notebook(
...
         preprocessor = PapermillExecutePreprocessor(
-            timeout=execution_timeout,
+            timeout=timeout,
             startup_timeout=start_timeout,
             kernel_name=kernel_name,
             log=logger,
+            **safe_kwargs
         )
         preprocessor.log_output = log_output
         preprocessor.preprocess(nb_man, kwargs)
```
captainsafia: Nit: I know it is not relevant to this invocation but can you pass `safe_kwargs` here instead of `kwargs`?

vizhur:
@captainsafia yup, will update preprocess call as well. Thanks for reviewing

MSeal:
Merging, will open an issue about general kwarg usage.

PR #356, opened 2019-05-06
Update CONTRIBUTING.md for pip 19

wdhorton:
An additional flag is required for pip 19 to do the editable install
because papermill has a pyproject.toml file.


MSeal:
Thanks!

Issue #354, opened 2019-05-05
How should we pass / control kwargs ?


MSeal:
In https://github.com/nteract/papermill/pull/351 it was raised that the manner in which we pass kwargs hasn't been well defined. We have several places where we pass everything downstream with **kwargs. But we don't control for duplicate names, especially as we're making API improvements.

Having kwarg passing is very useful from plugins we didn't write to be able to control their execution easily, and I'd prefer we keep a mechanism for passing parameters and data to any registered handler methods to support use cases we haven't thought or which are specific to a particular use case.

Off the top of mind I can think of two simple ways to address the name overlap issue.

- Make a function wrapper to safely map kwargs, prioritizing function def names over kwargs that overlap
- Always convert kwargs to safe_kwargs dict we pass without the `**` splat pattern

One of us will need to implement the solution we land on, likely before the next release. I assigned maintainers involved in the original discussion for now.

betatim:
Some things I think don't need to be solved by papermill are situations like:
```python
def foo(a, **kwargs):
    pass

foo(1, a=3)
```
If you end up in a place where this is a problem that needs solving I think it is a sign that we need to backup and take a different direction somewhere earlier. If this happens the solution is documentation (I think).

I think for passing additional (keyword) arguments to engines we should change `execute_notebook` to:
```
def execute_notebook(
    input_path,
    output_path,
...
    cwd=None,
    extra_engine_args=None,  # or maybe an empty dict
):
```
which users can then call as `execute_notebook(..., extra_engine_args=dict(foo='baz', whiz='bang'))`. This makes it clear where to put these extra arguments and that papermill will treat this as an opaque thing that it "blindly" passes on. Along the lines of "use this as an escape hatch, we hope you checked with the engine first that it knows what to do with these. You are voiding your papermill warranty by doing this.".

There is one wrinkle: some arguments need translating (`execution_timeout` <-> `timeout`, `start_timeout` <-> `startup_timeout`, etc). My hunch is that if a user sets an argument in their `extra_engine_args` then they want it to "win". At least I'd be super confused if I set something in an "for experts" bit of the API and it doesn't seem to make a difference.

[...]

PR #360, opened 2019-05-08
control kwargs

vizhur:
adding utils to handle kwargs to start from


willingc:
Hi @vizhur. Thanks for working on Papermill. Can you explain why you are making this PR? What is the use case? Why this would be a good approach to address the use case? This will help us evaluate the PR since I don't believe it closes any open issue and would be a new feature request.

vizhur:
@willingc that is regarding https://github.com/nteract/papermill/issues/354
related to the recent changes https://github.com/nteract/papermill/pull/351

papermill/utils.py
```diff
@@ -1,11 +1,31 @@
...
+import warnings
 
 from contextlib import contextmanager
 from functools import wraps
 
 logger = logging.getLogger('papermill.utils')
 
+WARN_ON_ARGS_OVERWRITE = True
+
+def safe_kwargs(caller_args, **callee_args):
```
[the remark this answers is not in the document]
MSeal: Probably needs a docstring to explain how it is meant to be used for future developers.

papermill/utils.py
```diff
@@ -1,11 +1,31 @@
...
+        conflicts = set(caller_args) & set(callee_args)
+        if conflicts:
+            args = format('; '.join(['{}={}'.format(key, value)
+                                     for key, value in callee_args.items()]))
+            msg = 'Callee will overwrite caller''s argument(s): {}'.format(args)
+            warnings.warn(msg)
+    return dict(caller_args, **callee_args)
+
+
+def remove_args(args=None, kwargs=None):
```
MSeal: `def remove_resvered_args(reserved_args=None, user_kwargs=None):`

papermill/utils.py
```diff
@@ -1,11 +1,31 @@
...
+
+def safe_kwargs(caller_args, **callee_args):
+    if WARN_ON_ARGS_OVERWRITE:
+        conflicts = set(caller_args) & set(callee_args)
+        if conflicts:
+            args = format('; '.join(['{}={}'.format(key, value)
+                                     for key, value in callee_args.items()]))
+            msg = 'Callee will overwrite caller''s argument(s): {}'.format(args)
+            warnings.warn(msg)
+    return dict(caller_args, **callee_args)
```
MSeal: Is the intention to overwrite application set argument with user arguments or the other way around?

papermill/utils.py
MSeal: Is this the inverse of `safe_kwargs`?

papermill/utils.py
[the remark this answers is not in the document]
vizhur: @MSeal if it happens down the stream that can be not user args anymore... It's generic a bit, but it's a util function that doesn't 'suppose' to be called by a user. Anyway, once we agreed on the solution I would be happy to add docstrings

papermill/utils.py
```diff
@@ -1,11 +1,31 @@
...
 
 from contextlib import contextmanager
 from functools import wraps
 
 logger = logging.getLogger('papermill.utils')
 
+WARN_ON_ARGS_OVERWRITE = True
+
+def safe_kwargs(caller_args, **callee_args):
+    if WARN_ON_ARGS_OVERWRITE:
```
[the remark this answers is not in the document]
vizhur: there were concerns about warnings.. I'd like to be able to turn off them for various reasons

willingc:
@MSeal In [Python docs](https://docs.python.org/3/reference/compound_stmts.html#function-definitions):

> If the form “**identifier” is present, it is initialized to a new ordered mapping receiving any excess keyword arguments, defaulting to a new empty mapping of the same type. 

I'm still not sure what case we are trying to prevent?? I don't understand why we can't just pass the kwargs.

papermill/utils.py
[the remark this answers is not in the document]
MSeal: Fair. Names were just confusing enough I couldn't reason about what actor is taking precedence.

papermill/utils.py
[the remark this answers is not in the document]
MSeal: Users can catch (or use contexts) to suppress warnings. Python actually has a good built-in control plane for them without us.

If you do want a control switch beyond that it shouldn't be a const in the file as editing that from outside the package would be a bit of an anti-pattern.

MSeal:
@willingc in the simplest form this happens now:

```
def top_level_func(foo, bar=None, **kwargs):
  downstream_func(bar=bar, timeout=50, **kwargs) # TypeError: downstream_func() got multiple values for keyword argument 'timeout'
```

We pass kwargs to multiple downstream callers which get augmented with named defaults which can conflict. I think @betatim 's main complaint was that it's making implicit contracts about reserved keywords in the kwargs. @vizhur was making an attempt to generate a utility to warn when these keywords conflict and continue execution rather than raising.

The fact it's about how we should handle implicit internal contracts when exposing `kwargs` and that we hadn't made a decision about how we'd handle these conflicts meant I pushed for #351 to merge as is and to discuss outside if we wanted to be more explicit in our handling of the problem. One option is to continue with passing **kwargs and let it raise when there's a conflict and not merge anything. But the issue had opinions that we should be a little more aware of potential kwargs conflicts in our plugin system as version bumps occur.

Does that makes the PR context a little clearer?

willingc:
Thank @MSeal for the clarification. I need to do a bit of thinking about this and also PEP 570. I'm trying to balance whether we are overly constraining which over time could be a maintenance headache against being more explicit about behavior (avoiding @betatim's implicit contracts about reserved keywords).

MSeal:
@vizhur Ok got enough of a direction to move forward. Outside of the single-line comments, let's do this:

- Let's change `safe_kwargs` to `merge_kwargs` to make it clearer what the function goal is.
- Then let's add it to the few places where we pass kwargs down and combine with other arguments for the next function call.
- Name kwargs to be the clearer about where these args are being passed (e.g. `engine_kwargs`)
- Add some tests for emitting warnings on merging overriding kwargs (I can help here if setting up the mocks / warnings capture is difficult) 
- Finally we should add to the DEVELOPMENT_GUIDE.md about minimizing passing extra kwargs downstream to functions.

Once the PR has these we can do normal PR review and get a merge + release.

papermill/utils.py
vizhur: function can overwrite arguments it passes down

papermill/utils.py
vizhur: kinda, need to remove args from kwargs intended for funtion but not listed in kwargs or named differently, like timeout/execution_timeout

vizhur:
@MSeal @willingc please review the changes

papermill/tests/test_engines.py
```diff
@@ -1,520 +1,520 @@
-import abc
```
MSeal: Do you know why this file got entirely touched? Was there a newline swap?

MSeal:
Outside of the one file being completely touched this LGTM. Given it's compressed to a few easily read lines of code I'm not too worried about maintenance / complexity burden here. Unless Carol objects or wants any changes I'll plan to merge and release after the test_engines.py comment.

vizhur:
@MSeal I reverted test_engines mess

willingc:
@vizhur @MSeal I appreciate the work that you are doing on this PR and the prior #351. I'm going to approve the PR but wish to open an issue to look at the design around engines and execution. Down the road, we may wish to refactor, but for now this seems to meet Microsoft's needs so let's go ahead and merge.

As next steps, I plan to document in more detail what is happening under the hood #368. After that, we should do a design review re: engines and execution to see what could be improved / clarified.

Thanks and :shipit:

willingc:
Thanks @vizhur :cake: Always great to have new contributors. Welcome!

PR #378, opened 2019-06-12
fix ADLHandler registration in papermill_io

VincentBLortie:
I should update more often, this has been broken for a few months 😰


MSeal:
Yikes! Thanks for fixing

PR #380, opened 2019-06-13
Moves imports to be lazy which pull in IPython

MSeal:
This saves substantially on initial load times for papermill, and reduces initial execution latency if progress bars are turned off.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,10 +3,6 @@
 """"
 setup.py
 
-Note: Do a version check for IPython.
-    IPython v6+ no longer supports Python 2.
-    If Python 2, install ipython 5.x.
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -67,30 +63,6 @@ def read_reqs(fname):
...
-            )
-    except Exception:
-        # We only want to optimistically report old versions
-        pass
-
-    if pip_message:
-        print(pip_message, file=sys.stderr)
-        sys.exit(1)
-
-
```
rgbkrk: This was for an old version of pip and previously needing IPython on start, so it makes sense to remove this. 👍

Issue #386, opened 2019-06-26
Missing instructions for JupyterLab usage


duarteocarmo:
I think we could add an explanation on how to add tags in JupyterLab. 

This could be particularly useful for folks now moving to it. 

For your reference, this is done by editing the metadata section in JupyterLab: 

![image](https://user-images.githubusercontent.com/26342344/60173245-93e91180-980e-11e9-8c0a-82b2c8ab5e53.png)

I suggest we add a note in the documentation. What do you think?

rgbkrk:
We'd love that addition to the documentation!

duarteocarmo:
Will make a pull request

duarteocarmo:
@rgbkrk  just to confirm, since there is no particular documentation specific to contributing to docs. 

My guess is I can clone repo and follow [this](https://travis-ci.org/nteract/papermill/jobs/550623562)?

basically run 
```bash
$ tox -e docs
```
right?

rgbkrk:
Yeah, let me know how that goes and perhaps you can also [add a section to `CONTRIBUTING.md` for docs](https://github.com/nteract/papermill/blob/master/CONTRIBUTING.md).

duarteocarmo:
Just a note, after installing the environment specified by doing `conda create -n dev`, and trying to preview the `.rst` files, the [`recommonmark`](https://pypi.org/project/recommonmark/) package was missing. (it's imported over at [`conf.py`](https://github.com/nteract/papermill/blob/master/docs/conf.py#L25))

rgbkrk:
Hmmm, seems we need it added to https://github.com/nteract/papermill/blob/master/requirements-dev.txt

duarteocarmo:
Just created a first pull request with the parameterization instructions, what do you think?

duarteocarmo:
@rgbkrk I also added a section on how to generate docs and test them. See #390 😄 

Going to sleep now, I'm all done. Cool project btw, great work

PR #389, opened 2019-06-26
Adds recommonmark dependency for building sphinx docs

duarteocarmo:
Adds [recommonmark](https://github.com/rtfd/recommonmark) dependency to [`requirements-dev.txt`](https://github.com/nteract/papermill/blob/master/requirements-dev.txt). Requested and referenced in #388 right [here](https://github.com/nteract/papermill/issues/386#issuecomment-506017765).

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -18,6 +18,7 @@ ipywidgets
 flake8
 tox
 bumpversion
+recommonmark
 pip>=18.1
 wheel>=0.31.0
 setuptools>=38.6.0


rgbkrk:
Thank you! 

@willingc this came up as something necessary in order to build/run docs locally. Let me know if our approach needs adjustment.

willingc:
Hi @duarteocarmo! Welcome. I always love it when a new contributor comes on board and is interested in docs.

The PR is fine and others may find helpful to have it in this dev requirements file too.

We do use tox to build the docs which pulls the [requirements file from the docs directory](https://github.com/nteract/papermill/blob/master/tox.ini#L21) (which is also what RTD is getting its requirements from.

PR #394, opened 2019-07-01
Matlab translation

RoelBrouwer:
Adding a MatlabTranslator to translations.py

This was created for a project where Matlab notebooks had to be parameterized. If you would like to extend language support to Matlab, this is what you are looking for :)

For those who want to review the choices made in translating Python concepts to Matlab:
  * **None**. Matlab does not have an exact None equivalent. I chose to use ``[]``, which, combined with a check using ``isempty(var)``, provides the best alternative in my opinion. ``NaN`` and ``isNaN(var)`` would be another option.
  * **string**. Double-quoted strings (") are the most close translation of Python strings. Single quoted (') strings exist as well, but are defined as character arrays.
  * **dictionary**. ``container.Map`` is similar to a Python dictionary. It maps keys (character arrays) to values. Note that a character array (using single quotes ') is different from a double-quoted (") string in Matlab. I implemented a separate function to translate the keys correctly. Another alternative would be ``struct``, a struct has fields (accessed using dot notation) that can contain any value. However, ``container.Map`` mirrors the concept of dictionaries more closely.
  * **list**. For lists I used cell arrays. A cell array is a data type with indexed data containers called cells, where each cell can contain any type of data.

In the codify method, I added a semicolon after each assignment, to suppress output. Otherwise, the notebook would print the result of each of the assignments in the (injected) parameters cell.


RoelBrouwer:
``papermill/tests/test_cli.py::test_pipe_out_auto`` fails (some of the time, but not consistently). Any ideas?

MSeal:
Thanks for putting this PR together!

For feature side, I think `None` -> `NaN` is probably a better translation than `[]`, albeit it's not perfect either. I agree with your assertions around the other data types and semi-colon.

The test_pipe_out_auto failure is odd. I don't think it's related to your PR in anyway. I'll investigate that failure independently. We can merge even if that's still plaguing this PR's builds.

papermill/tests/test_translators.py
```diff
@@ -238,6 +238,62 @@ def test_translate_comment_julia(test_input, expected):
...
 
 
+@pytest.mark.parametrize(
+    "test_input,expected",
+    [
+        ("foo", '"foo"'),
+        ('{"foo": "bar"}', '"{""foo"": ""bar""}"'),
+        ({"foo": "bar"}, 'containers.Map({\'foo\'}, {"bar"})'),
+        ({"foo": '"bar"'}, 'containers.Map({\'foo\'}, {"""bar"""})'),
+        ({"foo": ["bar"]}, 'containers.Map({\'foo\'}, {{"bar"}})'),
```
MSeal: Let's add a test for `{1: "foobar"}` to make sure the translator does something sane (I think it should translate the 1 to a string if the matlab Map can't support it as an integer).

MSeal: Same for `{1.0: "foobar"}`, {None, "foobar"}, and "{True: "foobar"} since Matlab has tighter type constraints here than other languages.

papermill/translators.py
```diff
@@ -213,12 +213,65 @@ def comment(cls, cmt_str):
...
+        return '"{}"'.format(str_val)
+
+    @staticmethod
+    def __translate_char_array(str_val):
+        """Translates a string to a Matlab char array"""
+        if isinstance(str_val, string_types):
+            str_val = str_val.encode('unicode_escape')
+            if sys.version_info >= (3, 0):
+                str_val = str_val.decode('utf-8')
+            str_val = str_val.replace('\'', '\'\'')
```
MSeal: This won't catch `\\'` to correctly leave it escaped, but that's probably an edge we don't need to solve for now.

papermill/tests/test_translators.py
RoelBrouwer: Good point. I have added these tests. They pass, but you might want to review if the current implementation displays the behavior that you would expect. Currently, it simply makes a char-array out of everything, so 1.0 becomes '1.0', None becomes 'None' and True becomes 'True'.

RoelBrouwer:
> For feature side, I think `None` -> `NaN` is probably a better translation than `[]`, albeit it's not perfect either. I agree with your assertions around the other data types and semi-colon.

Changed in commit 024797428d457b4f38e8c242d0b3990c22fbd8b8. In the end I think it comes down to personal preference, there is no perfect solution here.

papermill/translators.py
RoelBrouwer: Do you have a suggestion how to handle this?

RoelBrouwer:
> The test_pipe_out_auto failure is odd. I don't think it's related to your PR in anyway. I'll investigate that failure independently. We can merge even if that's still plaguing this PR's builds.

The builds for my last two commits did pass without problems, so the issue might have mysteriously disappeared...

papermill/translators.py
MSeal: If we needed to we would probably do `str_val.replace('\'', '\'\'').replace('\'', '\'\'')` in order to capture escaped string, or use a regex pattern. This is fine thinking about it more, as the original user input shouldn't be doing any pre-escaping patterns for this ever.

papermill/tests/test_translators.py
MSeal: That's reasonable. Thanks for adding the tests

RoelBrouwer:
I put WIP in because of the test_pipe_out_auto that failed, but that is no longer an issue. So you are right in assuming that there is no work in progress anymore :)

MSeal:
Perfect. Thanks for the contribution!

Issue #392, opened 2019-06-30
Option for execute_notebook() to disable expansion of parameters in the input path


0anton:
Dear papermilll maintainers,

Is it possible to have an option for execute_notebook(), which disables expansion of parameters in the source path?

I am using parameters in the file name to automatically produce multiple reports stemming from the same source notebook, e.g.:

```
papermill.execute_notebook('{param}/notebook.ipynb', 'value1/notebook.ipynb',
                                parameters={'param': 'value1'})
papermill.execute_notebook('{param}/notebook.ipynb', 'value2/notebook.ipynb',
                                parameters={'param': 'value2'})
```
                                
I cannot use this pattern with the current version of papermill, because it expands the input notebook name producing:

```
value1/notebook.ipynb -> value1/notebook.ipynb 
value2/notebook.ipynb -> value2/notebook.ipynb 
```

This fails because the source notebook does not exists.

I'd like to have all target reports produced from the single template notebook, which includes parameters in the path:
```
{param}/notebook.ipynb -> value1/notebook.ipynb 
{param}/notebook.ipynb -> value2/notebook.ipynb 
```

I cannot do it, because papermill expands parameters in both source and target file.

My suggestion is to introduce a bool option _parametrize_input_path_ in _execute_notebook()_, which disables expansion of parameters in the source path, like:

```
papermill.execute_notebook('{param}/notebook.ipynb', 'value1/notebook.ipynb',
                                parameters={'param': 'value1'}, 
                                parametrize_input_path=False)
```

I include suggested PR to the issue.

Kind regards,
Anton

MSeal:
Sorry for the delayed response.

I'm a little confused why your source notebook would have a `{param}` literal in the path, specifically the `{` symbols. It seems like your input path could easily be calculated before you pass it into papermill if you needed to apply some parameter ahead of time (with a `.format` call). If you really need `{param}` exactly in your input path you can escape the `{` and `}` characters by doing `{{param}}` instead.

I'm somewhat against adding this option, not because it wouldn't work, but because it complicates the papermill interface (including the cli). I would  Is there something I am missing about the input path here?

0anton:
Hi Matthew,

[...]

PR #397, opened 2019-07-08
 add option to avoid parametrization of input path

0anton:


0anton:
This is to solve #392.

MSeal:
Going to close this as the parent issue was solved by changing the call pattern.

Issue #363, opened 2019-05-10
Provide a list parameter in the command line


Ezibenroc:
I would like to pass a list as parameter value.
Currently, this can be done with a yaml file:
```yaml
my_list:
    - abc
    - def
    - ghi
```
Or the yaml code can be passed in the command line:
```shell
papermill input.ipynb output.ipynb -y '{my_list:[abc,def,ghi]}'
```
However, I would like to achieve such behavior with something like this (a list of values separated by spaces):
```shell
papermill input.ipynb output.ipynb -p my_list abc def ghi
```

----

My use case is that I often have notebooks that load a bunch of CSV files into a large pandas dataframe. I would like to use file globbing like this:
```shell
papermill input.ipynb output.ipynb -p my_files some/path/*.csv another/path/*.csv
```

MSeal:
That sounds convenient. I'm not 100% sure how to make click allow for 1-* values for a key, but I'm sure it'd be possible to add with a little exploration.

MSeal:
Closing this issue as a won't-do scenario, see #441 for details.

sumitbinnani:
@MSeal now that we have "-r" as an option, is list supported for CLI?

PR #411, opened 2019-08-14
list parameter

gshiba:
Hello!

This PR handles #363 

```
papermill input.ipynb output.ipynb -p my_list abc def ghi
```

But, it arguably relies on hacking `click.Option`, and `click` authors seem to be against variadic options [1].

Full credit goes to user stephen-rauch on stackoverflow [2]

The `class OptionEatAll` should probably be placed somewhere better...

[1] https://github.com/pallets/click/pull/259
[2] https://stackoverflow.com/a/48394004

### DIFF
--- a/papermill/cli.py
+++ b/papermill/cli.py
@@ -39,7 +39,7 @@ def print_papermill_version(ctx, param, value):
 @click.argument('notebook_path', required=not INPUT_PIPED)
 @click.argument('output_path', required=not (INPUT_PIPED or OUTPUT_PIPED))
 @click.option(
-    '--parameters', '-p', nargs=2, multiple=True, help='Parameters to pass to the parameters cell.'
+    '--parameters', '-p', nargs=-1, multiple=True, help='Parameters to pass to the parameters cell.'
 )
 @click.option(
     '--parameters_raw', '-r', nargs=2, multiple=True, help='Parameters to be read as raw string.'
@@ -171,8 +171,13 @@ def papermill(
         parameters_final.update(read_yaml_file(files))
     for params in parameters_yaml or []:
         parameters_final.update(yaml.load(params, Loader=NoDatesSafeLoader))
-    for name, value in parameters or []:
-        parameters_final[name] = _resolve_type(value)
+    for name, *values in parameters or []:
+        if not values:
+            raise ValueError
+        elif len(values) == 1:
+            parameters_final[name] = _resolve_type(values[0])
+        else:
+            parameters_final[name] = [_resolve_type(v) for v in values]
     for name, value in parameters_raw or []:
         parameters_final[name] = value
 
--- a/papermill/cli.py
+++ b/papermill/cli.py
@@ -218,21 +218,17 @@ def papermill(
     for params in parameters_yaml or []:
         parameters_final.update(yaml.load(params, Loader=NoDatesSafeLoader))
     for name_values in parameters or []:
-        if len(name_values) == 1:
-            raise ValueError("There must be at least one value for key '{}'".format(name_values[0]))
         name, values = name_values[0], name_values[1:]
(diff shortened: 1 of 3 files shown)


MSeal:
Thanks for the PR! This change seems solid, despite the nasty click hacks required. Let me fix the failing test (it's failing on master too) and make play around with this change some before merging.

gshiba:
The `-r/--raw` option should probably get the same treatment?

papermill/cli.py
```diff
@@ -171,10 +217,20 @@ def papermill(
         parameters_final.update(read_yaml_file(files))
     for params in parameters_yaml or []:
         parameters_final.update(yaml.load(params, Loader=NoDatesSafeLoader))
-    for name, value in parameters or []:
-        parameters_final[name] = _resolve_type(value)
-    for name, value in parameters_raw or []:
-        parameters_final[name] = value
+    for name, *values in parameters or []:
```
MSeal: This isn't legal syntax in python 2 (which we're still supporting through the end of the year)

gshiba:
Thanks for reviewing!

I hopefully fixed the py2 compatibility issues -- tests are still failing though.

Ok with where I placed the `class OptionEatAll`?

papermill/cli.py
```diff
@@ -171,10 +217,22 @@ def papermill(
         parameters_final.update(read_yaml_file(files))
     for params in parameters_yaml or []:
         parameters_final.update(yaml.load(params, Loader=NoDatesSafeLoader))
-    for name, value in parameters or []:
-        parameters_final[name] = _resolve_type(value)
-    for name, value in parameters_raw or []:
-        parameters_final[name] = value
+    for name_values in parameters or []:
+        name, values = name_values.pop(0), name_values
```
MSeal: I _think_ this is causing the test failures.
```suggestion
        name, values = name_values[0], name_values[1:]
```

papermill/cli.py
```diff
@@ -171,10 +217,22 @@ def papermill(
...
+        if not values:
+            raise ValueError("There must be at least one value for key '{}'".format(name))
+        elif len(values) == 1:
+            parameters_final[name] = _resolve_type(values[0])
+        else:
+            parameters_final[name] = [_resolve_type(v) for v in values]
+    for name_values in parameters_raw or []:
+        name, values = name_values.pop(0), name_values
```
MSeal:
```suggestion
        name, values = name_values[0], name_values[1:]
```

papermill/cli.py
```diff
@@ -171,10 +217,22 @@ def papermill(
...
+        if not values:
+            raise ValueError("There must be at least one value for key '{}'".format(name))
+        elif len(values) == 1:
+            parameters_final[name] = values[0]
+        else:
+            parameters_final[name] = values
```
MSeal:
```suggestion
            parameters_final[name] = list(values)
```

MSeal:
I'm good with where you placed OptionEatAll as it's not so large it needs its own file and relates to the problem immediately below it.

I think with the suggested change this will pass and I can add it to the release about to go out.

papermill/cli.py
gshiba: Hmmm, interesting that they may be the cause.

`name_values[1:]` could raise `IndexError`...

MSeal: Nope it's safe:

```
In [1]: t = []                                                                                                                              
In [2]: t[1:]                                                                                                                               
Out[2]: []
```

gshiba: Ah.. thanks.

gshiba: Ah, I get it now. `name_values` is a `tuple`, not a `list`.

MSeal:
Thanks for the improvement and working through the test failures!

PR #422, opened 2019-09-04
Add stdout-file/stderr-file options

akx:
This complements the existing log_output option to make it possible to more directly stream the notebook's stdout/stderr messages into separate files, without Papermill's internal log messages interfering.

Conveniently, `--stdout-file /dev/stdout --stderr-file /dev/stderr` lets one connect the two streams directly to their respective files.

All that is implemented in the last commit – there's a bunch of smaller preparatory commits that made it more convenient to develop this.

### DIFF
--- a/papermill/preprocess.py
+++ b/papermill/preprocess.py
@@ -2,7 +2,7 @@ from __future__ import unicode_literals, print_function
 
 from nbconvert.preprocessors import ExecutePreprocessor
 from nbconvert.preprocessors.execute import CellExecutionError
-from traitlets import Bool
+from traitlets import Bool, Instance
 
 
 class PapermillExecutePreprocessor(ExecutePreprocessor):
@@ -10,6 +10,8 @@ class PapermillExecutePreprocessor(ExecutePreprocessor):
     and updates outputs"""
 
     log_output = Bool(False).tag(config=True)
+    stdout_file = Instance(object, default_value=None).tag(config=True)
+    stderr_file = Instance(object, default_value=None).tag(config=True)
 
     def preprocess(self, nb_man, resources, km=None):
         """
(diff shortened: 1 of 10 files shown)


papermill/preprocess.py
```diff
@@ -67,20 +69,33 @@ def papermill_process(self, nb_man, resources):
...
+        :return:
+        """
         if output.output_type == "stream":
+            content = "".join(output.text)
             if output.name == "stdout":
-                self.log.info("".join(output.text))
+                if self.log_output:
+                    self.log.info(content)
+                if self.stdout_file:
+                    self.stdout_file.write(content)
```
MSeal: You'll want to do `.encode('utf-8')` if it's python 2 (python 3 is fine as is).

You can see the issue by adding this to a cell:
```
print('😀')
```

papermill/preprocess.py
```diff
@@ -67,20 +69,33 @@ def papermill_process(self, nb_man, resources):
...
+                    self.stdout_file.flush()
             elif output.name == "stderr":
-                # In case users want to redirect stderr differently, pipe to warning
-                self.log.warning("".join(output.text))
-        elif "data" in output and "text/plain" in output.data:
+                if self.log_output:
+                    # In case users want to redirect stderr differently, pipe to warning
+                    self.log.warning(content)
+                if self.stderr_file:
+                    self.stderr_file.write(content)
```
MSeal: Same deal with encoding when in python 2

akx:
@MSeal Thank you for the kind words!

I fixed the issue you brought up (it's been too long since I've had to deal with Python 2, I'd almost forgotten the horrors) by adding the `encoding='utf-8'` parameter to the `click` options instead:

https://github.com/nteract/papermill/pull/422/commits/6bc615a7c77f2d82f8d59db558409405df6c506d#diff-93e91e1558bb9064f6fda224500ec901R143

That way people who might be using Papermill as a library can pass in a file-like object that accepts text strings for `write` instead.

MSeal:
Great, thanks for fixing that issue. And the python 2 woes will start to drop off soon I hope. Papermill is going to drop support for it at the end of the year when it is officially unsupported by python.

PR #424, opened 2019-09-06
Remove CLI tests that don't actually test the CLI

akx:
These two tests only basically test that `click` actually passes values through.

I think they might have been some vestigial experiment code.


MSeal:
Good catch. Yes they generally read that way.

Issue #427, opened 2019-09-07
`-p` arguments right before notebook paths incorrectly uses notebook paths as option arguemnts


MSeal:
Related to #417 where the more complex option parsing is causing a different issue.

`papermill -p foo bar papermill/tests/notebooks/simple_execute.ipynb /tmp/out.ipynb` reproduces the issue.

MSeal:
Fixed in 1.2.0

Issue #417, opened 2019-08-23
papermill 1.1.0 attempts to parse second arguments for `-p` and `-r`


MSeal:
The following command since the release fails:

`papermill papermill/tests/notebooks/simple_execute.ipynb /tmp/test.ipynb -p 'command_line_args' '-from_date=20190723'`

as a result of the smarter multi-argument parsing of those options. Fix should be to ignore all click parsing for the second argument to capture at-least-one value for the parameter.

MSeal:
Fixed in 1.2.0

PR #428, opened 2019-09-07
Revert "list parameter (#411)"

MSeal:
This reverts commit 30c8ad1596a4850fea77d9a1cb3cba13e8a279c5.

@gshiba I found a fix for #417 but in in making that fix I discovered that there's no sane way to avoid capturing the notebook path when the last argument before the notebook input/outputs is a `-p`. We can't tell the `-p` option how many arguements there are and there's no way to detect if the notebook path is meant to be an additional list argument or a notebook path. The handler for notebook paths is definable by external users and has a wide range of acceptable values, so checking if the path is a notebook during option parsing is challenging to the point of impossibility for all use-cases.

See #427's example:

`papermill -p foo bar papermill/tests/notebooks/simple_execute.ipynb /tmp/out.ipynb`

After giving this some thought I think the only path forward is to revert the change. Additionally since the more complex option parsing caused two bugs that weren't obvious from review I think maybe the fancier option parsing is not a good idea for long-term maintenance of the repo.

We'll have to keep using the `-y '{"foo": ["bar", "baz"]}' pattern for passing list arguments to papermill.

### DIFF
--- a/docs/usage-execute.rst
+++ b/docs/usage-execute.rst
@@ -61,23 +61,6 @@ remain a string, even if it may be interpreted as a number or boolean.
 
     $ papermill local/input.ipynb s3://bkt/output.ipynb -r version 1.0
 
-Setting a list of values for a parameter
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-Both ``-p`` and ``-r`` support multiple values, which will be turned into a list. For example:
-
-.. code-block:: bash
-
-   $ papermill input.ipynb output.ipynb -p foo bar baz 42 -p piyo hoge -r spam ham eggs 45
-
-will populate the parameter cell with:
-
-.. code-block:: python
-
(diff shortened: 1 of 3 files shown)


gshiba:
Fair points!

I also encountered an 'issue' with the possibly multi value option, which is that I needed to check if the value is a list or not in the notebook...

```python
# Parameters cell
files = ['file1', file2']  # CLI: papermill in.ipynb out.ipynb -p files file1 file2
# or
files = 'file1'  # CLI: papermill in.ipynb out.ipynb -p files file1

# next cell
if not is instance(files, list):
    files = [files]
```

I was thinking of suggesting a separate option (e.g. `-l`) to explicitly specify that the 1+ values will be parsed into a list, but you'd still need some separator (e.g. `--`) to separate the options and arguments (which `click` doesn't support).

willingc:
Thanks @gshiba and @MSeal for working together on this. I'm going to go ahead and merge the PR.

Issue #420, opened 2019-09-03
Piping papermill into nbconvert triggers an encoding error on Python 2


mwouts:
For some notebooks (possibly those with non-ascii characters in the output cells, I will have to confirm), piping papermill into nbconvert triggers an encoding error (stacktrace below).

Maybe we could replace

https://github.com/nteract/papermill/blob/01ec212f14c76860f3788f6f92f9be27d5742ea3/papermill/iorw.py#L104

with

```python
return sys.stdout.write(buf.encode('utf-8'))
```

The substitution above fixes my issue locally, but possibly we need to do this conditionally on the type of `buf`.  And maybe a similar fix on the standard input would be appropriate (see also #317). 

I have not been able to extract a MRE yet, but the stacktrace is
```
Traceback (most recent call last):
  File "/home/mwouts/conda/envs/jp2/bin/papermill", line 10, in <module>
    sys.exit(papermill())
  File "/home/mwouts/conda/envs/jp2/lib/python2.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/mwouts/conda/envs/jp2/lib/python2.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/mwouts/conda/envs/jp2/lib/python2.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/mwouts/conda/envs/jp2/lib/python2.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/mwouts/conda/envs/jp2/lib/python2.7/site-packages/papermill/cli.py", line 254, in papermill
    cwd=cwd,
  File "/home/mwouts/conda/envs/jp2/lib/python2.7/site-packages/papermill/execute.py", line 107, in execute_notebook
    write_ipynb(nb, output_path)
  File "/home/mwouts/conda/envs/jp2/lib/python2.7/site-packages/papermill/iorw.py", line 363, in write_ipynb
    papermill_io.write(nbformat.writes(nb), path)
  File "/home/mwouts/conda/envs/jp2/lib/python2.7/site-packages/papermill/iorw.py", line 104, in write
    return sys.stdout.write(buf)
```

MSeal:
[...]

PR #430, opened 2019-09-11
Explicit utf-8 encoding on stdout

mwouts:
This implements the change suggested by @MSeal at #420.

I confirm that this solves the original issue with my notebook, as well as the one with `echo 'import world_bank_data as wb; wb.get_countries()' | jupytext --from py --to ipynb --set-kernel - | papermill > out.ipynb`.

@MSeal , I had to modify the test `test_write_stdout`. It is expected that I had to mock `stdout` with `io.BytesIO` instead of `io.StringIO`?

### DIFF
--- a/papermill/iorw.py
+++ b/papermill/iorw.py
@@ -101,7 +101,13 @@ class PapermillIO(object):
 
     def write(self, buf, path, extensions=['.ipynb', '.json']):
         if path == '-':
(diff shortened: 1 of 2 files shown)


MSeal:
@mwouts Thanks for the fix! Yes I think the change to io.BytesIO is correct.

PR #441, opened 2019-11-02
Support list parameters (#363)

### LINKED ISSUE #363: Provide a list parameter in the command line

Raab70:
Closes #363 

This PR adds support for list parameters in the CLI. The current implementation supports comma-separated lists. We could of course change this. Also note I did not update documentation as I was a bit unsure. Does this feature deserve it's own section [here](https://github.com/nteract/papermill/blob/master/docs/usage-execute.rst#L39)?

I added a test around the CLI parsing. It likely does not cover the other small change which was to catch `TypeError`'s which were thrown when calling `int(list_)`

I was also able to test this with:

`papermill papermill/tests/notebooks/simple_execute.ipynb test.ipynb -p msg test,list,message` and parameters are passed into the notebook correctly.

### DIFF
--- a/papermill/tests/test_cli.py
+++ b/papermill/tests/test_cli.py
@@ -119,6 +119,15 @@ class TestCLI(unittest.TestCase):
         )
 
     @patch(cli.__name__ + '.execute_notebook')
+    def test_parameters(self, execute_patch):
+        self.runner.invoke(
+            papermill, self.default_args + ['-p', 'foo', 'bar,baz', '--parameters', 'buzz', '42']
+        )
+        execute_patch.assert_called_with(
+            **self.augment_execute_kwargs(parameters={'foo': ['bar', 'baz'], 'buzz': 42})
+        )
+
+    @patch(cli.__name__ + '.execute_notebook')
     def test_parameters_raw(self, execute_patch):
         self.runner.invoke(
             papermill, self.default_args + ['-r', 'foo', 'bar', '--parameters_raw', 'baz', '42']
--- a/requirements-azure.txt
+++ b/requirements-azure.txt
@@ -1,3 +1,3 @@
 azure-datalake-store >= 0.0.30
-azure-storage-blob
+azure-storage-blob ~= 2.1
 requests >= 2.21.0
--- a/papermill/cli.py
+++ b/papermill/cli.py
@@ -29,7 +29,7 @@ class ListOrString(click.types.StringParamType):
     sep = ','
 
     def convert(self, value, param, ctx):
-        val = super().convert(value, param, ctx)
+        val = super(ListOrString, self).convert(value, param, ctx)
         split_val = val.split(self.sep)
         if len(split_val) > 1:
             return split_val
--- a/papermill/cli.py
+++ b/papermill/cli.py
@@ -24,6 +24,18 @@ INPUT_PIPED = S_ISFIFO(os.fstat(0).st_mode)
 OUTPUT_PIPED = not sys.stdout.isatty()
 
 
+class ListOrString(click.types.StringParamType):
+    name = 'list_or_string'
+    sep = ','
+
+    def convert(self, value, param, ctx):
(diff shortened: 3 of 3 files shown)


Raab70:
I saw this same error locally and it has nothing to do with my changes:

```
E   ImportError: cannot import name 'BlockBlobService' from 'azure.storage.blob' (/home/vsts/work/1/s/.tox/py37/lib/python3.7/site-packages/azure/storage/blob/__init__.py)
```

When I look at the `azure-storage-blob` package on pypi there is a very strange 12.x release in the last few days ([here](https://pypi.org/project/azure-storage-blob/#history)). This would have been after the last master build. 

However, the latest commit coincides with the 2.1.0 release back in August: https://github.com/Azure/azure-storage-python/tree/master/azure-storage-blob

I'm not sure if some Azure maintainer's credentials were compromised but I changed the requirements to avoid pulling in the new broken release.

papermill/tests/test_cli.py
```diff
@@ -118,6 +118,15 @@ def test_parameters(self, execute_patch):
             **self.augment_execute_kwargs(parameters={'foo': 'bar', 'baz': 42})
         )
 
+    @patch(cli.__name__ + '.execute_notebook')
+    def test_parameters(self, execute_patch):
+        self.runner.invoke(
+            papermill, self.default_args + ['-p', 'foo', 'bar,baz', '--parameters', 'buzz', '42']
```
MSeal: So I'm a little hesitant on this feature change. An earlier attempt https://github.com/nteract/papermill/pull/411 had to be reverted as it broke other CLI interactions. Specifically my concern for this PR is that many users pass JSON as parameter inputs: e.g. `-p '["foo","bar"]'`. Another issue would be that this would currently block users from passing "bar,baz" if they needed to as a string. It might be such that we want to not approach this problem in the CLI layer itself and leave yaml/json inputs to handle list/map types. Open to conversation though if you see a clean way past concerns.

requirements-azure.txt
```diff
@@ -1,3 +1,3 @@
 azure-datalake-store >= 0.0.30
-azure-storage-blob
+azure-storage-blob ~= 2.1
```
MSeal: Thanks for identifying this. I'll look into this fix for the upstream changes independently as well/

papermill/tests/test_cli.py
Raab70: Totally understand the concerns. My thinking was definitely that the YAML/JSON inputs would be better suited for inputting lists. I was just going off the ticket. I don't really think there's a clean way to do this without breaking those use cases. The only thought would be adding a flag to enable/disable this functionality but I wouldn't exactly say that's clean. That seems like a confusing workaround.

MSeal: Yeah, I missed closing the original issue when reverted the other attempt. It can get pretty gnarley to handle all the cases correctly while maintaining a simple interface with `-p`options (if it's even possible to do so).

Raab70: @MSeal Maybe some additional examples in the documentation on how to pass in different types of parameters would help users? I could add a section specifically on "complex types" focusing on the options for passing in lists/structs.

MSeal: That'd be awesome, if you wanted to make a docs PR!

PR #448, opened 2019-11-23
Translator for C# (.net-csharp) kernel

lobrien:
Translator for the recently-released .net-csharp kernel. 

## Pre-requisites

- Dotnet core 3.0 and 2.1 (https://dotnet.microsoft.com/download) 
- "Dotnet try" tool (https://www.nuget.org/packages/dotnet-try/) 
`dotnet tool install --global dotnet-try`

- .net kernels for Jupyter
`dotnet try jupyter install`

(See https://www.hanselman.com/blog/AnnouncingNETJupyterNotebooks.aspx )

## Notes

I had to lock the `azure-storage-blob` reference to the mid-2019 version, which is still supported. The Azure Python SDK was updated in late October with breaking changes. 

I didn't implement "None," although I think it might be possible to do something clever with `Nullable`s.

I implemented List as an array and Dict as a Dictionary<Object,Object?>. I could make them more consistent by making List `List<Object>` but that seems YAGNI.

### DIFF
--- a/papermill/tests/test_translators.py
+++ b/papermill/tests/test_translators.py
@@ -194,16 +194,13 @@ def test_translate_codify_scala(parameters, expected):
         ("foo", '"foo"'),
         ('{"foo": "bar"}', '"{\\"foo\\": \\"bar\\"}"'),
         ({"foo": "bar"},
-            'new Dictionary<string,Object>(new [] '
-            + '{ new KeyValuePair<string,Object>("foo" , "bar") })'),
+            'new Dictionary<string,Object>{ { "foo" , "bar" } }'),
         ({"foo": '"bar"'},
-            'new Dictionary<string,Object>(new [] '
-            + '{ new KeyValuePair<string,Object>("foo" , "\\"bar\\"") })'),
+            'new Dictionary<string,Object>{ { "foo" , "\\"bar\\"" } }'),
         (["foo"], 'new [] { "foo" }'),
         (["foo", '"bar"'], 'new [] { "foo", "\\"bar\\"" }'),
         ([{"foo": "bar"}],
-            'new [] { new Dictionary<string,Object>('
-            + 'new [] { new KeyValuePair<string,Object>("foo" , "bar") }) }'),
+            'new [] { new Dictionary<string,Object>{ { "foo" , "bar" } } }'),
         (12345, '12345'),
         (-54321, '-54321'),
         (1.2345, '1.2345'),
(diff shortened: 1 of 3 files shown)


papermill/translators.py
```diff
@@ -272,13 +272,55 @@ def codify(cls, parameters):
...
+
+    @classmethod
+    def translate_dict(cls, val):
+        """Translate dicts to nontyped dictionary"""
+
+        kvps = ', '.join(
+            ["new KeyValuePair<Object,Object>({} , {})".format(cls.translate_str(k), cls.translate(v))
+             for k, v in val.items()]
+        )
+        return 'new Dictionary<Object,Object?>(new [] {{ {} }})'.format(kvps)
```
MSeal: I think you could make this `Dictionary<string,Object?>` safely to help with key typing at least.

papermill/tests/test_translators.py
```diff
@@ -187,6 +187,73 @@ def test_translate_codify_scala(parameters, expected):
...
+@pytest.mark.parametrize(
+    "parameters,expected",
+    [
+        ({"foo": "bar"}, '// Parameters\nvar foo = "bar";\n'),
+        ({"foo": True}, '// Parameters\nvar foo = true;\n'),
+        ({"foo": 5}, '// Parameters\nvar foo = 5;\n'),
+        ({"foo": 1.1}, '// Parameters\nvar foo = 1.1;\n'),
+        ({"foo": ['bar', 'baz']}, '// Parameters\nvar foo = new [] { "bar", "baz" };\n'),
+        ({"foo": {'bar': 'baz'}},
+            '// Parameters\nvar foo = new Dictionary<Object,Object?>(new [] '
```
MSeal: I struggle to get this to compile. Is there a specific version of C# this works (can't test the .net kernel specifically from this computer atm)?

Seems like `var foo = new Dictionary<Object,Object?> { { "bar", "baz" } }\n` would work (and be a little easier to read)?

MSeal: Do we need `using System.Collections.Generic;` for this to work in the kernel? We could inject that into the parameter cell if it is needed.

requirements-azure.txt
```diff
@@ -1,3 +1,3 @@
 azure-datalake-store >= 0.0.30
-azure-storage-blob
-requests >= 2.21.0
+azure-storage-blob == 2.1.0
+requests >= 2.1.0
```
MSeal: typo?

requirements-azure.txt
```diff
@@ -1,3 +1,3 @@
 azure-datalake-store >= 0.0.30
-azure-storage-blob
-requests >= 2.21.0
+azure-storage-blob == 2.1.0
```
MSeal: Thanks -- someone needs to dig into why the incompatible change in a separate PR . But pinning for now is probably the right change.

papermill/tests/test_translators.py
lobrien: ... still working on this ...

lobrien: Re. `using`: YAGNI, apparently. I confirmed that it works without explicitly bringing that in.

MSeal: Ok, great -- just figured I'd check for the kernel. The setup must be providing some basic things on behalf of the user.

requirements-azure.txt
MSeal: I think this is the last thing to check on before merge

MSeal:
Awesome, thanks for adding support for the kernel!

This will probably go out with the papermill 2.0 (probably releasing around the end of the year) unless you need this sooner in a released version.

PR #449, opened 2019-11-30
Ending notebook execution early with sys.exit without raising an exception

### LINKED ISSUE #435: a good way to early stop the script
Hi!

is there a way to "exit" the execution properly?
sys.exit() throws an exception

yo-my-bard:
Hi!

I'm a newbie around these parts (and contributing to open source in general!) but thought I might make an attempt to resolve #435 . Noticed that without arguments to exit() or with None, `evalue` returns an empty string; that should also be evaluated the same as 0 I imagine? Any and all feedback is welcome 🙏🏾

Thank you!
Yusuph

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -168,6 +168,8 @@ def raise_for_execution_errors(nb, output_path):
 
         for output in cell.outputs:
             if output.output_type == "error":
+                if output.ename == "SystemExit" and (output.evalue == "" or output.evalue == "0"):
+                    continue
                 error = PapermillExecutionError(
                     exec_count=cell.execution_count,
                     source=cell.source,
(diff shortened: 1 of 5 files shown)


MSeal:
The change looks good, we just need a couple unittests to cover the behavior. Probably one unittest with an `exit()`, `exit(0)`, and finally an `exit(1)` added to `test_execute.py`.  Don't worry about the build failing in travis / azure, there's an issue with azure blob dependency causing it to fail atm.

yo-my-bard:
Thanks for the review! Added tests -- tried to follow convention but let me know if I need to make any changes.

Also, this is strange but when I was running my tests, I occasionally got this in the output: `Timeout waiting for IOPub output` . It would cause that test to fail. When that did not happen, the tests would pass. Not sure if that's related to the new tests or not.

PR #451, opened 2019-12-07
Translator: F# (dotnet-fsharp)

lobrien:
See PR https://github.com/nteract/papermill/pull/448 for prerequisites / install

### DIFF
--- a/papermill/translators.py
+++ b/papermill/translators.py
@@ -312,6 +312,42 @@ class CSharpTranslator(Translator) :
     def assign(cls, name, str_val):
         return 'var {} = {};'.format(name, str_val)
 
+class FSharpTranslator(Translator) :
+
+    @classmethod
+    def translate_none(cls, val) :
+        return 'None'
+
+    @classmethod
+    def translate_bool(cls, val) :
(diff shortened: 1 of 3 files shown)


papermill/translators.py
```diff
@@ -313,6 +313,43 @@ def assign(cls, name, str_val):
...
+
+    @classmethod
+    def translate_int(cls, val):
+        strval = cls.translate_raw_str(val)
+        return strval + "L" if (val > 2147483647 or val < -2147483648) else strval
+
+    @classmethod
+    def translate_dict(cls, val):
+        tuples = '; '.join(
+            ["({}, {} :> IComparable)".format(cls.translate_str(k), cls.translate(v))
```
MSeal: Haven't done much F#, so not 100% sure this is the right / best option but I think it's a reasonable and working pattern.

MSeal:
Thanks for the PR!

papermill/translators.py
lobrien: Yeah, I'm pretty sure this is sub-elegant!

PR #455, opened 2019-12-12
Use Black to format Python parameters for versions >= 3.6

### LINKED ISSUE #436: PrettyPrint parameters
When we inject parameters they form very large single lines if the parameter is a map or list of arguments. It's be nice for readability if the parameter injection was smart about formatting inputs and ran them through a linter or printed maps / lists in a way that was more legible. This might become language / kernel specific, but would still be a solid improvement to generated notebooks.

petebachant:
Resolves #436.

I picked cutoff values for when to use newlines/indents based on my own practices. Was thinking of trying to limit line length, but we don't know the variable name so we can't calculate the full line length.

### DIFF
--- a/papermill/translators.py
+++ b/papermill/translators.py
@@ -121,15 +121,40 @@ class PythonTranslator(Translator):
 
     @classmethod
     def translate_dict(cls, val):
-        escaped = ', '.join(
-            ["{}: {}".format(cls.translate_str(k), cls.translate(v)) for k, v in val.items()]
+        # If dict has more than 3 items, use hanging indents
+        if len(val) > 3:
+            sepchar = '\n'
+            indent = '    '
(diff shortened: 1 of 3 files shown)


MSeal:
IMO I think importing and reusing https://github.com/psf/black/blob/master/black.py would do better here. Choosing split by element count will still look ugly with nested objects that are really long, or one long string.

Also for testing, we'll probably need to either run the test results through the same formatter function, or pass a flag to not format for most tests and separately test for just the format impact.

petebachant:
@MSeal the tests are actually all passing even when run through the Black formatter.

MSeal:
I'll try to test this out tonight. Thanks for getting the PR up!

MSeal:
Work great, merging for the next major release :)

Issue #454, opened 2019-12-11
Proposal: Save notebook file in the middle of long cell executions


leopd:
(Apologies if this has come up before - I looked for an issue about it, but didn't find one.)

Similar to #433 I also run long machine learning jobs in notebooks, where single cells can take hours or even days to run, so I want to see the cell's intermediate results before that cell is completed.  I know that `--log-output` will print it to the console, but I'd like a feature where the `.ipynb` file gets updated periodically during long executions.

I think I see a reasonable way to do this in the code.  Proof of concept at https://github.com/nteract/papermill/compare/master...leopd:save-mid-cell
Clearly this is incomplete as it should be configurable, and maybe even have some smarts to detect if saving the notebook periodically is taking too long since this is an intrinsically N^2 operation that could slow down over time with large notebooks and cloud storage.  Happy to work on that stuff, but wanted to get some feedback first.

Has this been considered/rejected before?  Does it seem worthwhile?  Any side-effects or subtleties I should think about?

MSeal:
There's been a few questions on "when does my notebook get updated". Most of those cases were really about checking current status, which log-outputs solves for users. However people's initial reaction is always go to the notebook and it's convenient when that's up-to-date during long executions. I think you hit on most of the risks with more frequent saves, namely large files or storage system rate limits. As long as it's configurable I don't see a serious problem with adding some default saving rate into papermill, similar to the linked implementation. If you wanted to make a PR we can bundle that functionality along with the papermill 2.0 release for python 3 only.

leopd:
Submitted a PR, so closing this.

PR #458, opened 2019-12-20
Autosaves notebook during long cell executions. 

leopd:
Following up proposal in #454 - this (auto-)saves the notebook in the middle of long cell executions.  It's CLI-configurable, and has some decent end-to-end tests.  Couldn't get tox to run "binder" on my machine, but everything else is looking good from here.

I'll highlight some design trade-off decisions I made that I'm not attached to at all if y'all have better ideas:

- Is the name "autosave" right for this feature?  I think it's fairly self-explanatory and well understood, and matches reasonably closely what's going on.  But it's not super similar to the feature it's most closely related to "request-save-on-cell-execute" which IMHO is somewhat of an awkward name. (Who are we requesting this from?)
- The default behavior: autosave every 30 seconds during long cell executions.
- The back-off behavior.  It's set to automatically double the autosave time if it every takes more than 25% of the autosave delay to do the write-to-disk.  (e.g. if it takes >7.5 seconds to save the notebook when autosave is every 30s)  This seemed to me like a reasonable behavior to protect against cases where the notebook is very big, or the storage is very slow.
- Configuration of back-off behavior: there currently is none.  The 25% is fixed unless you instantiate the objects yourself.  The trade-off here is control vs simplicity.  I figured that the complexity of explaining to everybody what's going on, and making the CLI options bigger, just wasn't worth the additional control.  I thought about exposing the percent, and a max time, but it all seemed like a lot of words for everybody to read that they'd almost never use.  Again, totally open to revisiting that.

Since y'all had mentioned that this would probably go into a py3-only release, I was tempted to include some type hints for clarity, but didn't because that would leave the style inconsistent and require changing tox.ini.


papermill/tests/test_autosave.py
```diff
@@ -0,0 +1,58 @@
...
+            time.sleep(0.55)  # Sleep for long enough that autosave should work, if enabled
+            nb_man.autosave_cell()
+            assert write_mock.call_count == 0  # but it's disabled.
+
+    def test_end2end_autosave_slow_notebook(self):
+        test_dir = tempfile.mkdtemp()
+        nb_test_executed_fname = os.path.join(test_dir, 'output_{}'.format(self.notebook_name))
+        with patch.object(engines, 'write_ipynb') as write_mock:
+            # This notebook has a cell which takes 2.5 seconds to run
+            execute_notebook(self.notebook_path, nb_test_executed_fname, autosave_cell_ever=1)
```
leopd: Gah.  That's not right.  `autosave_cell_ever` should be `_every`

leopd: This test was passing, but should not have been.  Default saves is actually 4.  Rewrote it.

PR #459, opened 2019-12-23
Add --execution_timeout flag for failing on too-slow cells

huonw:
This updates the CLI docs to include the full output of `--help` after
this patch; it seems that some flags have been added without
regenerating it, so the diff includes more than just this new
`--execution_timeout` flag.

### DIFF
--- a/papermill/cli.py
+++ b/papermill/cli.py
@@ -118,10 +118,14 @@ def print_papermill_version(ctx, param, value):
     help='Set log level',
 )
 @click.option(
-    '--start_timeout', type=int, default=60, help="Time in seconds to wait for kernel to start."
+    '--start-timeout',
+    '--start_timeout',  # Backwards compatible naming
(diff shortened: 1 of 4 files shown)


papermill/cli.py
```diff
@@ -111,6 +111,11 @@ def print_papermill_version(ctx, param, value):
 @click.option(
     '--start_timeout', type=int, default=60, help="Time in seconds to wait for kernel to start."
 )
+@click.option(
+    '--execution_timeout',
```
MSeal: This should be `--execute-timeout` for the cli (click will convert it to `execution_timeout` automatically)

MSeal: Since no one was looking at PRs for a while over the holidays, I'll make this change now and get this merged now

huonw: Thanks 👍

PR #480, opened 2020-03-20
Use a markdown cell for error flagging, prettier than %%html

huonw:
Markdown is (mostly) a superset of HTML, meaning there's a lot of HTML one can include directly as markdown content. Papermill can use this to have a slightly cleaner error alert at the top of a failed notebook, by not having both code and output, just output only.

This makes the rendering of a broken notebook (such as `papermill/tests/notebooks/broken.ipynb`) cleaner in an interactive Jupyter instance, as well as on the GitHub and NBViewer static renderers:

Before:
- <https://gist.github.com/huonw/b08434842b6b9f6d45090141ae4327dd>
- <https://nbviewer.jupyter.org/gist/huonw/b08434842b6b9f6d45090141ae4327dd>
  image

After:
- <https://gist.github.com/huonw/f22a6caf6bf8c31538000b88e28c05da>
- <https://nbviewer.jupyter.org/gist/huonw/f22a6caf6bf8c31538000b88e28c05da>
  image

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -182,13 +182,7 @@ def raise_for_execution_errors(nb, output_path):
     if error:
         # Write notebook back out with the Error Message at the top of the Notebook.
         error_msg = ERROR_MESSAGE_TEMPLATE % str(error.exec_count)
-        error_msg_cell = nbformat.v4.new_code_cell(
-            source="%%html\n" + error_msg,
-            outputs=[
-                nbformat.v4.new_output(output_type="display_data", data={"text/html": error_msg})
-            ],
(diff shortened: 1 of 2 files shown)


MSeal:
This is a nice change! Thanks for making it. One of the tests caught that the cell type changed, you'll want to tweak that test some to better capture that the exception message was added to the top of the notebook output (likely need to change it to check the markdown content instead of cell type).

huonw:
Ah, oops! I must've missed it when I was updating tests locally.

MSeal:
Thanks!

PR #483, opened 2020-03-26
Include link to the cell with an error, in failure heading

huonw:
Papermill currently inserts an error message `An Exception was encountered at 'In [NNN]'` at the top of a notebook that fails to execute, where NNN is the cell execution count of the failing cell. This patch adjusts the `In [NNN]` text to be a link to the relevant cell, to make it easy to jump to. This works by inserting another markdown cell immediately before the failing cell, where the markdown cell contains an HTML anchor (specifically, `id="papermill-error-cell"`). The heading then links to it with a `#papermill-error-cell` fragment HREF.

This helps reduce the scrolling and searching required for long notebooks where the error cell might be one of dozens, or hidden among long output. For an example of the latter, see <https://nbviewer.jupyter.org/gist/huonw/dd13fa49b5876bfea26b69832702d573>.

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -144,14 +144,15 @@ def prepare_notebook_metadata(nb, input_path, output_path, report_mode=False):
     return nb
 
 
+ERROR_STYLE = 'style="color:red; font-family:Helvetica Neue, Helvetica, Arial, sans-serif; font-size:2em;"'
 ERROR_MESSAGE_TEMPLATE = (
-    '<span style="color:red; font-family:Helvetica Neue, Helvetica, Arial, sans-serif; font-size:2em;">'
+    '<span ' + ERROR_STYLE + '>'
     "An Exception was encountered at '<a href=\"#papermill-error-cell\">In [%s]</a>'."
     '</span>'
 )
 
 ERROR_ANCHOR_MSG = (
-    '<span id="papermill-error-cell" style="color:red; font-family:Helvetica Neue, Helvetica, Arial, sans-serif; font-size:2em;">'
+    '<span id="papermill-error-cell" ' + ERROR_STYLE + '>'
     'Papermill encountered exception here:'
     '</span>'
 )
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -83,6 +83,8 @@ def execute_notebook(
             nb = parameterize_notebook(nb, parameters, report_mode)
 
         nb = prepare_notebook_metadata(nb, input_path, output_path, report_mode)
+        # clear out any existing error markers from previous papermill runs
+        nb = remove_error_markers(nb)
 
         if not prepare_only:
             # Fetch the kernel name if it's not supplied
@@ -144,9 +146,12 @@ def prepare_notebook_metadata(nb, input_path, output_path, report_mode=False):
     return nb
 
 
+ERROR_MARKER_TAG = "papermill-error-cell-tag"
+
 ERROR_STYLE = (
     'style="color:red; font-family:Helvetica Neue, Helvetica, Arial, sans-serif; font-size:2em;"'
 )
+
 ERROR_MESSAGE_TEMPLATE = (
     '<span ' + ERROR_STYLE + '>'
     "An Exception was encountered at '<a href=\"#papermill-error-cell\">In [%s]</a>'."
@@ -158,6 +163,16 @@ ERROR_ANCHOR_MSG = (
     'Execution encountered an exception here and stopped:'
     '</span>'
 )
+
+
+def remove_error_markers(nb):
+    nb = copy.deepcopy(nb)
+    nb.cells = [
+        cell
+        for cell in nb.cells
+        if ERROR_MARKER_TAG not in cell.metadata.get("tags", [])
+    ]
+    return nb
 
 
 def raise_for_execution_errors(nb, output_path):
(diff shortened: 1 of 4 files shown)


papermill/execute.py
```diff
@@ -144,9 +144,18 @@ def prepare_notebook_metadata(nb, input_path, output_path, report_mode=False):
...
-    '<span style="color:red; font-family:Helvetica Neue, Helvetica, Arial, sans-serif; font-size:2em;">'
-    "An Exception was encountered at 'In [%s]'."
+    '<span ' + ERROR_STYLE + '>'
+    "An Exception was encountered at '<a href=\"#papermill-error-cell\">In [%s]</a>'."
+    '</span>'
+)
+
+ERROR_ANCHOR_MSG = (
+    '<span id="papermill-error-cell" ' + ERROR_STYLE + '>'
+    'Papermill encountered exception here:'
```
MSeal:
```suggestion
    'Execution encountered an exception here and stopped:'
```

huonw:
Yeah, that concern was why I chose to have the second cell have text in it, rather than just `<span id="papermill-error-cell"></span>`, so that it would be more obvious and more likely that people remember to remove it. I like the approach of tags and automatically remove them, when re-executed via Papermill. I'll whip up an implementation now 😄 

I also looked for a Jupyter-native way of linking to specific cells, but couldn't find one. Maybe I missed it?

MSeal:
There is not a Jupyter-native way to link to cells -- it's been a hot topic for debate which likely needs to have a cell id exposed to notebook in the next major format version bump. See https://github.com/jupyter/nbformat/issues/148

Another option is to inject an html output at the end of the cell before's `outputs` block with the link info. This would make rerunning the notebook in any interface clear the result.

huonw:
> I'll whip up an implementation now

Done.

> There is not a Jupyter-native way to link to cells -- it's been a hot topic for debate which likely needs to have a cell id exposed to notebook in the next major format version bump. See jupyter/nbformat#148

Ah, I see.

> Another option is to inject an html output at the end of the cell before's outputs block with the link info. This would make rerunning the notebook in any interface clear the result. Though this would appear in the prior cell's output section which might be visibly odd.

Ah, so manually edit the output? I think this won't work well if the cell(s) before the error are markdown cells and so don't have separate output?

MSeal:
> Ah, so manually edit the output? I think this won't work well if the cell(s) before the error are markdown cells and so don't have separate output?

Yep that's true :/

papermill/execute.py
```diff
@@ -144,13 +146,35 @@ def prepare_notebook_metadata(nb, input_path, output_path, report_mode=False):
...
-    '<span style="color:red; font-family:Helvetica Neue, Helvetica, Arial, sans-serif; font-size:2em;">'
-    "An Exception was encountered at 'In [%s]'."
+    '<span ' + ERROR_STYLE + '>'
+    "An Exception was encountered at '<a href=\"#papermill-error-cell\">In [%s]</a>'."
+    '</span>'
+)
+
+ERROR_ANCHOR_MSG = (
+    '<span id="papermill-error-cell" ' + ERROR_STYLE + '>'
+    'Execution encountered an exception here and stopped:'
```
captainsafia, line 163: Nit: I would update the error text to indicate that this error was encountered when running the notebook through papermill.

captainsafia:
This is a tricky one. The ideal implementation would store information in the metadata of the notebook and leave each UI to determine how to display this error to the user. While, IMO, that is the best from a design perspective, it's difficult to accomplish.

For compatibility across different UIS and implementation cost, this seems sensible.

Left a quick nit to update the error message so that users can track where it came from (papermill).

papermill/execute.py
huonw: I had it mention Papermill originally, but @MSeal suggested this version in https://github.com/nteract/papermill/pull/483#discussion_r401206867

What do you both think about something like the following?

```suggestion
    'Execution using Papermill encountered an exception here and stopped:'
```

MSeal: That works imo, I'm not too partial on exact wording.

captainsafia: That looks good!

MSeal:
I know this has been sitting for awhile -- I'll set some time to hash our if we want papermill to inject the link cell or not with the wider nteract group and get a decision after our next sync.

@willingc @rgbkrk fyi :point_up:

huonw:
Thanks; that would be great.

For a little more context, I'm excited by this tweak or something equivalent, because it'll put the icing on the cake for how [we](https://github.com/stellargraph/stellargraph) are checking notebooks on CI and making them conveniently viewable: https://medium.com/stellargraph/better-notebooks-through-ci-automatically-testing-documentation-for-graph-machine-learning-5789e277e597 ("Highlighting errors with Buildkite artifacts and annotations" section in particular).

rgbkrk:
I like the injected cell at first glance, even with the tradeoff of having to delete that cell later.

Also... kind of tangentially, I think it's ok for us to make cell IDs and stick them in cell metadata when run via papermill. Maybe the way to unblock the inertia on that effort is to start providing something UIs can work with even if it's not fully planned.

MSeal:
Alright from talking I think we're all in agreement with @rgbkrk 's comment above. Going to merge and snap a patch release for this.

MSeal:
Also for cell IDs... i'm going to make a JEP soon based on https://discourse.jupyter.org/t/annotating-jupyter-notebooks/2079/11

huonw:
Thanks for merging 👍 Hopefully it doesn't take long for there to be a better way to do this!

Issue #515, opened 2020-06-17
Not Implemented Error Coming from ZMQ/asyncio


crawftv:
Getting a not implemented error when running execute_notebook.
python 3.8
windows 10
virtualenv
A link to my project I am working on is here. https://github.com/crawftv/crawto. 
I tried running it without a separate jupyter instance running but that did not help.

```shell
papermill crawto.ipynb crawto1.ipynb
Input Notebook:  crawto.ipynb
Output Notebook: crawto1.ipynb
Executing:   0%|                                                                              | 0/13 [00:00<?, ?cell/s]
Traceback (most recent call last):
  File "C:\Users\cwcol\AppData\Local\Programs\Python\Python38\lib\runpy.py", line 193, in _run_module_as_main  
    return _run_code(code, main_globals, None,  
  File "C:\Users\cwcol\AppData\Local\Programs\Python\Python38\lib\runpy.py", line 86, in _run_code  
    exec(code, run_globals)
  File "E:\projects\crawto\.test-env\Scripts\papermill.exe\__main__.py", line 7, in <module>  
  File "e:\projects\crawto\.test-env\lib\site-packages\click\core.py", line 829, in __call__  
    return self.main(*args, **kwargs)  
  File "e:\projects\crawto\.test-env\lib\site-packages\click\core.py", line 782, in main  
    rv = self.invoke(ctx)  
  File "e:\projects\crawto\.test-env\lib\site-packages\click\core.py", line 1066, in invoke  
    return ctx.invoke(self.callback, **ctx.params)  
  File "e:\projects\crawto\.test-env\lib\site-packages\click\core.py", line 610, in invoke  
    return callback(*args, **kwargs)  
  File "e:\projects\crawto\.test-env\lib\site-packages\papermill\cli.py", line 222, in papermill  
    execute_notebook(  
  File "e:\projects\crawto\.test-env\lib\site-packages\papermill\execute.py", line 95, in  
execute_notebook  
    nb = papermill_engines.execute_notebook_with_engine(  
  File "e:\projects\crawto\.test-env\lib\site-packages\papermill\engines.py", line 49, in  execute_notebook_with_engine  
    return self.get_engine(engine_name).execute_notebook(nb, kernel_name, **kwargs)  
  File "e:\projects\crawto\.test-env\lib\site-packages\papermill\engines.py", line 343, in execute_notebook    
    cls.execute_managed_notebook(nb_man, kernel_name, log_output=log_output, **kwargs)  
[...]

PR #517, opened 2020-06-24
Added guard for asyncio registration for python 38 on windows

MSeal:
Fix for #515.


MSeal:
Codecov complained because it's 1 line more uncovered -- tests are all passing.

PR #158, opened 2018-07-28
Display inferred params with `papermill foo.ipynb --help`

thomasballinger:
Working on guessing param type from a parameters cell

plan: mirror the structure of `_build_parameter_code` for different languages

Issue #55, opened 2017-10-25
Use parameter types to convert arguments


betatim:
Right now all arguments passed as `-p foo 2` are created as `float`s in the notebook. When using parameters to specify shapes of numpy arrays and the like you end up having to convert them to integers first. We could use the type of the value of the parameter as a guide and convert arguments to that type.

So a "parameters" cell with `foo = 2.123` and `papermill ... -p foo 2` would result in `2.` but a cell with `foo = 2` would result in `2`.

mpacer:
There could be some utilities for inspecting python type annotations for any non-third party code.

betatim:
#58 fixes a bug that stopped you from getting an `int` passed as parameter, but I still think trying to deduce the type from the type of the parameter's value in the notebook would be cool.

thomasballinger:
some related work in https://github.com/nteract/papermill/pull/158, displaying inferred types and default values with `--help`

PR #531, opened 2020-08-26
Guess params by introspecting the _parameters_ cell

### LINKED ISSUE #225: Introspect Parameters before Execution
## Question
I was curious if there was a way to introspect the default parameters of a Notebook before execution. I see that if I `execute_notebook` with parameters I can then read these afterwards, but I can't ask what they are before-hand.

## Implementation
Would it be appropriate to catch this during the `load_notebook_node` function? We could capture this as a `"default_parameters"` in the `nb.metadata`? Happy to put in a PR if this seems useful to anyone else.

https://github.com/nteract/papermill/blob/81af3f3807674d4ca651a66a29eb9b095a411da7/papermill/iorw.py#L213

fcollonval:
This is a reboot of #158 

It adds a new option `--help-notebook` to create a new cli usage:

```
papermill --help-notebook input_notebook.ipynb
```
Due to that I had to remove the `required` statement on OUTPUT_PATH. But the test is still done within the `papermill` function to keep the behavior if `--help-notebook` is not set.

Fixes #225 => in particular the default parameters resulting of the code introspection are added as a dictionary in metadata `nb.metadata['papermill']['default_parameters']`.

For now this add the ability for Python. I'm not familiar to the other languages to build the proper regex.

> Note: I use a regex instead of `ast` for Python for two reasons: `ast` removes all comments but I use them to get a help comment. ast parsing for non-python language won't be available. So it seems better to use regex for all languages.

A follow up could be to address #55, but there is a important issue on how to fairly evaluate the variable type for non-python language.

### DIFF
--- a/papermill/inspection.py
+++ b/papermill/inspection.py
@@ -1,14 +1,10 @@
- # -*- coding: utf-8 -*-
+# -*- coding: utf-8 -*-
 """
 Deduce parameters of a notebook from the parameters cell.
 """
-
-import ast
-
 import click
 
 from .iorw import get_pretty_path, open_notebook
-from .models import Parameter
 from .utils import any_tagged_cell
 
 
@@ -18,7 +14,7 @@ def display_notebook_help(notebook_path):
     the same automatic error messaging, but with the ability to hijack
     the normal help messaging for when a user types `--help` on an
     input notebook
-    
+
     Parameters
     ----------
     notebook_path : str
@@ -28,7 +24,7 @@ def display_notebook_help(notebook_path):
     pretty_path = get_pretty_path(notebook_path)
     click.echo("Usage: papermill [OPTIONS] {} [OUTPUT_PATH]".format(pretty_path))
     click.echo("\n  Parameters inferred for notebook '{}':".format(pretty_path))
-    
+
     if not any_tagged_cell(nb, "parameters"):
         click.echo("\n  No cell tagged 'parameters'")
         return
--- a/papermill/inspection.py
+++ b/papermill/inspection.py
@@ -30,8 +30,12 @@ def display_notebook_help(ctx, notebook_path):
     params = nb.metadata['papermill']['default_parameters']
     if params:
         for p in params.values():
-            type_repr = p["inferred_type_name"] if p["inferred_type_name"] != "None" else "Unknown type"
-            click.echo("     {}: {} (default {})\t\t{}".format(p["name"], type_repr, p["default"], p["help"]))
+            type_repr = p["inferred_type_name"]
+            if type_repr == "None":
+                type_repr = "Unknown type"
+            click.echo(
+                "     {}: {} (default {})\t\t{}".format(p["name"], type_repr, p["default"], p["help"])
+            )
     else:
         click.echo(
             "\n  Can't infer anything about this notebook's parameters. "
--- a/papermill/inspection.py
+++ b/papermill/inspection.py
@@ -29,8 +29,11 @@ def display_notebook_help(ctx, notebook_path):
 
     params = nb.metadata['papermill']['default_parameters']
     if params:
-        for p in params:
-            type_repr = p.inferred_type_name if p.inferred_type_name != "None" else "Unknown type"
-            click.echo("     {}: {} (default {})\t\t{}".format(p.name, type_repr, p.default, p.help))
+        for p in params.values():
+            type_repr = p["inferred_type_name"] if p["inferred_type_name"] != "None" else "Unknown type"
+            click.echo("     {}: {} (default {})\t\t{}".format(p["name"], type_repr, p["default"], p["help"]))
     else:
-        click.echo("\n  Can't infer anything about this notebook's parameters. It may not have any parameter defined.")
+        click.echo(
+            "\n  Can't infer anything about this notebook's parameters. "
+            "It may not have any parameter defined."
+        )
--- a/papermill/inspection.py
+++ b/papermill/inspection.py
@@ -8,21 +8,19 @@ from .iorw import get_pretty_path, open_notebook
 from .utils import any_tagged_cell
 
 
-def display_notebook_help(notebook_path):
-    """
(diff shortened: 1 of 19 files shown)


MSeal:
Thanks for posting the PR -- I just got back and will get a chance to go through it in the next day or two

papermill/iorw.py
```diff
@@ -422,6 +424,22 @@ def load_notebook_node(notebook_path):
...
+    parameter_cell = nb.cells[parameter_cell_idx]
+    kernel_name = nb.metadata.kernelspec.name
+    language = nb.metadata.kernelspec.language
+
+    translator = papermill_translators.find_translator(kernel_name, language)
+    try:
+        params = translator.inspect(parameter_cell)
+        nb.metadata['papermill']['default_parameters'] = {p.name: p._asdict() for p in params}
+    except NotImplementedError:
+        logger.warning(
```
MSeal, line 439: This might get annoying having this print on each notebook execution for language without inspection support. I think we should only pull out default parameters on demand when `--help` is inquired to avoid A) unnecessary warnings and B) errors in inspection blocking execution paths.

papermill/translators.py
```diff
@@ -114,8 +120,36 @@ def codify(cls, parameters, comment='Parameters'):
...
+        List[Parameter]
+            A list of all parameters
+        """
+        raise NotImplementedError('parameters introspection not implemented for {}'.format(cls))
+
 
 class PythonTranslator(Translator):
+    # Pattern to capture parameters within cell input
+    PARAMETER_PATTERN = re.compile(
+        r"^(?P<target>\w[\w_]*)\s*(:\s*[\"']?(?P<annotation>\w[\w_\[\],\s]*)[\"']?\s*)?=\s*(?P<value>.*?)(\s*#\s*(type:\s*(?P<type_comment>[^\s]*)\s*)?(?P<help>.*))?$"  # noqa
```
MSeal, line 150: Really nice actually -- it even extracts the comment on the end of `foo = bar  # a comment`!

MSeal: The other way to do this for Python is to use the `ast` module (e.g. https://www.mattlayman.com/blog/2018/decipher-python-ast/) and extracting the assignments programmatically. However this won't work cross-language and could even hiccup with ipython specific capabilities with python kernels (e.g. `!run-something`). So I think regex is the KISS way to go.

papermill/translators.py
```diff
@@ -158,6 +192,84 @@ def codify(cls, parameters, comment='Parameters'):
...
+                    flat_string += line[:comment_pos].strip()
+                else:
+                    flat_string += line.strip()
+            if len(accumulator):
+                flat_string += accumulator[-1].strip()
+            return flat_string
+
+        grouped_variable = []
+        accumulator = []
+        for iline, line in enumerate(src.splitlines()):
```
MSeal, line 238: Interesting combination of operations to allow for the regex to pick up the help text. Might be worth a comment explaining why this code is here and why the regex alone isn't sufficient (I had to come back after reading it to get why).

papermill/tests/test_inspect.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+@pytest.fixture
+def click_context():
+    mock = MagicMock(spec=Context, command=MagicMock())
+    mock.command.get_usage.return_value = "Dummy usage"
+    return mock
+
+
+@pytest.mark.parametrize(
```
MSeal, line 24: Probably want more examples (maybe by supplying the parameter cell directly to the inspect call in a different test parameterization) with a variety of code arrangements to prove your comment compressing code all works as expected.

requirements-dev.txt
```diff
@@ -1,19 +1,23 @@
+azure-datalake-store >= 0.0.30
+azure-storage-blob >= 12.1.0
```
MSeal, line 2: I believe these should only be in the requirements-azure.txt file

MSeal (changes requested):
Thanks for building this out. There's some comments to address and probably a little bit of rearrangement needed to avoid code paths when executing but I think the core implementation idea is solid in approach.

papermill/inspection.py
```diff
@@ -0,0 +1,43 @@
...
+        Path to the notebook to be inspected
+    """
+    nb = open_notebook(notebook_path)
+    click.echo(ctx.command.get_usage(ctx))
+    pretty_path = get_pretty_path(notebook_path)
+    click.echo("\n  Parameters inferred for notebook '{}':".format(pretty_path))
+
+    if not any_tagged_cell(nb, "parameters"):
+        click.echo("\n  No cell tagged 'parameters'")
+        return
```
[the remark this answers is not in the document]
fcollonval: Change to return 1 in that case, 0 otherwise

papermill/iorw.py
fcollonval: The goal was to achieve the proposal of #225.

But I totally get your point.

papermill/tests/test_inspect.py
fcollonval: This is actually done in  papermill/tests/test_translators.py:test_inspect_python (at line 112)

papermill/translators.py
fcollonval: That is exactly my note in the PR description :wink: 

Note: I use a regex instead of ast for Python for two reasons: ast removes all comments but I use them to get a help comment. ast parsing for non-python language won't be available. So it seems better to use regex for all languages.

papermill/translators.py
fcollonval: I added a comment

requirements-dev.txt
fcollonval: Sorry for this one.

papermill/translators.py
MSeal: Yep just re-enforcing the idea

MSeal (approved):
Thanks for the improvement and the quick turn around on the review!

Issue #536, opened 2020-09-19
sys.exit isn't handled gracefully in papermill 2.1.3


ctb:
I'd like to stop the execution of a notebook under certain circumstances, and per 

https://papermill.readthedocs.io/en/latest/changelog.html#features

I think I should be able to execute `sys.exit(0)` or `raise SystemExit`, but that doesn't seem to work for me.

This code:

![Screen Shot 2020-09-19 at 11 14 39 AM](https://user-images.githubusercontent.com/51016/93686185-6ae31b00-fa69-11ea-95a6-7435f4f32839.png)

Yields:

![Screen Shot 2020-09-19 at 11 14 30 AM](https://user-images.githubusercontent.com/51016/93686186-6c144800-fa69-11ea-8422-25d25a5e5748.png)

What am I doing wrong? :) 

```
% pip list | grep -i paper
papermill                                 2.1.3
% pip list | grep -i notebook
notebook                                  6.0.3
```

willingc:
Hi @ctb :smile:

Hmm... We're testing `sys.exit(0)`, `sys.exit(1)`, and `sys.exit()` in the Papermill tests. 

This seems to work if I put in a try block:

```
try:
    sys.exit()
except:
    print('Hello')
finally:
    print('Yo')
```

outputting: Hello Yo

and

```
try:
    SystemExit
except:
    print('Hello')
finally:
    print('Yo')
```
Outputs: Yo

@MSeal When support for `sys.exit(0)` was added, were there any constraints on the use cases that would prevent the above?

ctb:
hi @willingc!! Fancy meeting you here!

Well... what should a bare `sys.exit`/`SystemExit` do? I was assuming it would stop the notebook at that cell, which is what I want to do.

MSeal:
A bare `sys.exit(0)` should be making the papermill process exit cleanly. I'll take a look this week at the PR and issue more in depth but can't look much today.

ctb:
thanks, much appreciated! and while I'm at it ...thanks for papermill :)

MSeal:
So @ctb are you just seeing the error message in the notebook or actually seeing the papermill process fail? When sys.exit occurs the kernel process running the notebook still fails with the SystemExit error, we just treat that exception (which is printed in the notebook) as not an error to papermill. Here

```
01:12:03 mseal ~ (venv:.py3local)  
$ papermill Untitled.ipynb Untitled-out.ipynb 
Input Notebook:  Untitled.ipynb
[...]

PR #538, opened 2020-09-20
add test for SystemExit

willingc:
Related to #536


MSeal (approved):
Covers the manual sysexit raise call, thanks for including

PR #544, opened 2020-10-30
Allow `pathlib.Path` paths

### LINKED ISSUE #543: Failure when a path is pathlib.Path
The Papermill Python API fails if either of the path arguments supplied to `execute_notebook` is a `pathlib.Path`. This is because [`parameterize_path` assumes that the path is always a `str`](https://github.com/nteract/papermill/blob/d5299b0f2705d7a9175c377aa2a2f812c83239e3/papermill/parameterize.py#L50). In order to support parameterized paths, Papermill probably has to continue work with `str`s internally, but it would be nice if Papermill did not fail when given a `Path`. A rasonable fix would be to convert each `Path` object to a `str` before continuing.

drhagen:
Convert `Path` to `str` when received. Fixes #543.

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -2,6 +2,7 @@
 
 import copy
 import nbformat
+from pathlib import Path
 
 from .log import logger
 from .exceptions import PapermillExecutionError
@@ -32,9 +33,9 @@ def execute_notebook(
 
     Parameters
     ----------
-    input_path : str
+    input_path : str or Path
         Path to input notebook
-    output_path : str
+    output_path : str or Path
         Path to save executed notebook
     parameters : dict, optional
         Arbitrary keyword arguments to pass to the notebook parameters
@@ -56,7 +57,7 @@ def execute_notebook(
         Duration in seconds to wait for kernel start-up
     report_mode : bool, optional
         Flag for whether or not to hide input.
-    cwd : str, optional
+    cwd : str or Path, optional
         Working directory to use when executing the notebook
     **kwargs
         Arbitrary keyword arguments to pass to the notebook engine
@@ -66,6 +67,13 @@ def execute_notebook(
     nb : NotebookNode
        Executed notebook object
     """
+    if isinstance(input_path, Path):
+        input_path = str(input_path)
+    if isinstance(output_path, Path):
+        output_path = str(output_path)
+    if isinstance(cwd, Path):
+        cwd = str(cwd)
+
     path_parameters = add_builtin_parameters(parameters)
     input_path = parameterize_path(input_path, path_parameters)
     output_path = parameterize_path(output_path, path_parameters)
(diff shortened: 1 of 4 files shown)


papermill/tests/test_inspect.py
```diff
@@ -52,7 +52,12 @@ def click_context():
     ],
 )
 def test_inspect_notebook(name, expected):
-    assert inspect_notebook(str(name)) == expected
+    assert inspect_notebook(name) == expected
+
+
+def test_str_path():
+    expected = {"msg": {"name": "msg", "inferred_type_name": "None", "default": "None", "help": ""}}
```
captainsafia, line 59: Nit: Can we fix the formatting on this value?

papermill/execute.py
```diff
@@ -56,7 +57,7 @@ def execute_notebook(
         Duration in seconds to wait for kernel start-up
     report_mode : bool, optional
         Flag for whether or not to hide input.
-    cwd : str, optional
+    cwd : str or Path, optional
```
captainsafia, line 60: @willingc Are there other public APIs we might want to make this modification to?

captainsafia (approved):
Looks good from my end.

Posted a query to Carol to get her thoughts on other APIs that might need to be modified.

papermill/tests/test_inspect.py
drhagen: Sure. This was copied verbatim from [higher in the file](https://github.com/nteract/papermill/blob/main/papermill/tests/test_inspect.py#L30). What would you like it changed to and would you like me to fix the formatting throughout the file to match?

papermill/execute.py
willingc: Let me take a closer look tomorrow. Thanks @drhagen.

papermill/execute.py
```diff
@@ -2,6 +2,7 @@
 
 import copy
 import nbformat
+from pathlib import Path
```
rgbkrk, line 5: I just double checked since I wasn't sure -- `pathlib` was introduced in 3.4 and we require 3.6+ now. 👍

papermill/execute.py
rgbkrk: Maybe `inspect_notebook` in `papermill/inspection.py`

rgbkrk: Nevermind, @drhagen got that below.

papermill/tests/test_inspect.py
rgbkrk: Surprisingly, this is exactly how black formats this dict. I think we have to leave it as is. 😫

captainsafia: I've been living in JavaScript land for too long, haven't I? All good then!

rgbkrk (approved):

PR #548, opened 2020-11-04
Update extending-entry-points.rst

koek67:
Using the correct class name of the sample custom engine


captainsafia (approved):
Good catch! @willingc Can you deploy the doc update?

willingc:
Updated doc is available: https://papermill.readthedocs.io/en/latest/extending-entry-points.html?highlight=CustomEngine

Thanks @koek67

PR #561, opened 2020-12-11
set base docs python to 3 in tox.ini

amcinnes87:


willingc:
Our current tox implementation for docs requires 3.8. Contributors may be using different recent versions of Python 3 than 3.8, and docs builds error with 3.6 and 3.7. This makes it easier for contributors to be successful.

willingc (approved):
Thanks @amcinnes87. It was fun pairing on this last night.

PR #565, opened 2021-01-11
Added cell description to tqdm

GiovanniCiampi:
Added possibility to set description of cells that will be injected to the tqdm.
To add description to a specific cell the user must add a string with this formatting in the desired cell:
`# papermill_description=description_of_cell`
When the execution will reach that cell, the description of the tqdm will be changed from `Executing x:` to `Executing description_of_cell`. 
If no cell contains the tag `papermill_description=`, the execution will be as before the implementation of this functionality.
If a cell containing the description tag is followed by cells that don't have the tag, the tqdm description is set to the last valid description.

I decided to implement this functionality because I think it is pretty useful with long notebooks that have very slow cells (i.e. data preparation and model training). An alternative can be to use logging, but it is way more verbose.
With this implementation you can simply add the 'Training' description to a cell and you know you are in the training process, or you can add 'Data_Preparation' and you know you are doing data preparation, moreover you see only the description of the current cell in the tqdm.
If a user doesn't like this functionality, he can simply ignore it.

I think it would be a good idea to add information about this feature in the documentation, but I would like to know first if this this is correctly implemented and perceived as a good addition from the authors/contributors.

### DIFF
--- a/papermill/engines.py
+++ b/papermill/engines.py
@@ -212,6 +212,11 @@ class NotebookExecutionManager(object):
         cell.metadata.papermill['start_time'] = self.now().isoformat()
         cell.metadata.papermill["status"] = self.RUNNING
         cell.metadata.papermill['exception'] = False
+        
+        # injects optional description of the current cell directly in the tqdm 
+        cell_description = self.get_cell_description(cell)
+        if cell_description is not None and hasattr(self, 'pbar') and self.pbar:
+            self.pbar.set_description(f"Executing {cell_description}")
 
         self.save()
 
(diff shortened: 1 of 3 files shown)


MSeal (approved):
Thanks for the addition, well get this into the next release

MSeal:
Overall the approach seems fine to me for an opt-in pattern. There's not a great programmatic pattern for fetching first comment in a cell outside of loading a full AST parser that's comment aware. This is simpler and less dependency requiring to achieve the same thing.

Issue #572, opened 2021-01-23
Matlab issue: codify() takes 2 positional arguments but 3 were given


cskaandorp:
Hi there, 

I've upgraded from 2.3.0 to 2.3.1 but I still have a problem with executing a notebook with a Matlab Kernel:

Traceback (most recent call last):
  File "/Users/casper/opt/anaconda3/bin/gitmiller", line 8, in <module>
    sys.exit(main())
  File "/Users/casper/opt/anaconda3/lib/python3.8/site-packages/gitmiller/command_line.py", line 181, in main
    get_from_github(**function_arguments)
  File "/Users/casper/opt/anaconda3/lib/python3.8/site-packages/gitmiller/command_line.py", line 132, in get_from_github
    pm.execute_notebook(
  File "/Users/casper/opt/anaconda3/lib/python3.8/site-packages/papermill/execute.py", line 91, in execute_notebook
    nb = parameterize_notebook(nb, parameters, report_mode)
  File "/Users/casper/opt/anaconda3/lib/python3.8/site-packages/papermill/parameterize.py", line 80, in parameterize_notebook
    param_content = translate_parameters(kernel_name, language, parameters, comment)
  File "/Users/casper/opt/anaconda3/lib/python3.8/site-packages/papermill/translators.py", line 554, in translate_parameters
    return papermill_translators.find_translator(kernel_name, language).codify(parameters, comment)
TypeError: codify() takes 2 positional arguments but 3 were given

I did have a quick look but I don't understand the error message: codify is called with two parameters, not 3. Weird.
Thanks for looking into it.

RoelBrouwer:
It seems that when the optional parameter `comment` was introduced in the `codify` function, the MatLab translator `codify` was not updated. The change was applied in the `Translator` class (line 117-118 of translators.py) and the `PythonTranslator` class (line 185-186), but not in the `MatlabTranslator` class (line 406-407). The `codify` function of the MatlabTranslator does not know what to do with the third argument (`comment`, after `cls` and `parameters`) and therefore errors.
I think this can be fixed quiet easily by adapting lines 406-407 in a way analogous to the change to lines 117-118. I currently do not have a local papermill installation, but if needed I can submit a pull request that should fix this in the next couple of days.

MSeal:
Thanks, a Pull Request would be greatly appreciated!

MSeal:
Closing since this was resolved in https://github.com/nteract/papermill/pull/574. Release with changes is going out today

PR #574, opened 2021-01-25
Add optional comment parameter to MatlabTranslator's codify

RoelBrouwer:
Added the optional comment parameter to the codify function in MatlabTranslator, like it has been in Translator and PythonTranslator before.

This should fix the issue reported in #572.


MSeal (approved):
Would love for a test of the codify call here since tests didn't catch that codify contract was mismatched, but happy to have the fix regardless.

PR #587, opened 2021-04-20
add basic gh-action and 3.9

willingc:


tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 skipsdist = true
-envlist = py{36,37,38}, flake8, dist, manifest, docs, binder
+envlist = py{36,37,38,39}, flake8, dist, manifest, docs, binder
```
MSeal, line 3: Been a while since we touched that list :joy:

PR #588, opened 2021-04-22
update parameterization instructions for lab v3 and beyond

wangjoshuah:
Make adding parameters a bit more clear. I was using Jupyer Labs v3 and could not find the double gear icon since the screenshot was for 2.0-2.2


MSeal (approved):
Thanks for the documentation improvement!

PR #613, opened 2021-06-06
flake8: unify tox & setup

Borda:
## What does this PR do?

unify the confusion in setup.cfg and in tox file to use the same


MSeal (approved):
Thanks

PR #614, opened 2021-06-06
Updated file read to not fail early with a boto empty file exception

### LINKED ISSUE #603: Empty parameters file in S3 raises a InvalidRange exception
Hi,

When passing S3 URIs to papermills, the S3Handler will always try to access the byte range `bytes=0-`. If the targeted object is empty though, this will result in the S3 client raising an exception.

At first, I thought it might be a bug in the S3 client, but it turns out it just implements [RFC2616#14.35.1](https://datatracker.ietf.org/doc/html/rfc2616#section-14.35.1) to the letter:

> If a syntactically valid byte-range-set includes at least one byte-range-spec **whose first-byte-pos is less than the current length of the entity-body**, or at least one suffix-byte-range-spec with a non-zero suffix-length, then the byte-range-set is satisfiable. Otherwise, the byte-range-set is unsatisfiable.

For an empty file, `0` (first-bytes-pos) is **equal** to the length of the object, hence the range is "unsatisfiable", and `botocore` correctly handles it:

> If the byte-range-set is unsatisfiable, the server SHOULD return a response with a status of 416 (Requested range not satisfiable).

A possible fix would be to handle empty objects in a special way.

The affected code: https://github.com/nteract/papermill/blob/a2c6a0b101d606b87c2ee571d682fb5d1fcf06db/papermill/s3.py#L292-L307

My guess is that this patch should be enough to make the exception go away and still have everything else working:
```patch
diff --git a/papermill/s3.py b/papermill/s3.py
index 7ce3f20..6311e7e 100644
--- a/papermill/s3.py
+++ b/papermill/s3.py
@@ -304,6 +304,9 @@ class S3(object):
                 elif size != obj.content_length:
                     raise AwsError('key size unexpectedly changed while reading')

+                if size == 0:
+                    return
+
                 r = obj.get(Range="bytes={}-".format(bytes_read))

                 try:
```

Cheers!

MSeal:
## What does this PR do?

Breaks out of the read loop in s3 cat function to avoid [RFC2616#14.35.1](https://datatracker.ietf.org/doc/html/rfc2616#section-14.35.1).

Fixes #603

### DIFF
--- a/papermill/s3.py
+++ b/papermill/s3.py
@@ -304,6 +304,12 @@ class S3(object):
                 elif size != obj.content_length:
                     raise AwsError('key size unexpectedly changed while reading')
 
+                # For an empty file, 0 (first-bytes-pos) is equal to the length of the object
+                # hence the range is "unsatisfiable", and botocore correctly handles it by
+                # raising an exception. We'd rather just return with empty file contents here.
+                if size == 0:
+                    break
+
                 r = obj.get(Range="bytes={}-".format(bytes_read))
 
                 try:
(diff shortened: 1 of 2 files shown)


willingc (approved):
Love the tests. Thanks @MSeal

PR #622, opened 2021-07-22
Read notebook from github

### LINKED ISSUE #556: New feature: Read input notebook from github
In my org, I store every notebook in github and tweak papermill to read notebook from github directly.
Because we use papermill heavily in production and we need to version notebooks.

This is how we use
```python
import papermill as pm

pm.execute_notebook(
   'https://github.com/nteract/papermill/blob/main/papermill/tests/notebooks/read_check.ipynb',
   'path/to/output.ipynb',
   parameters = dict(alpha=0.6, ratio=0.1)
)
```

Take just url of notebook like binder and nbviewer.
This has some pros.
Some teams use only dev / master branches.(Read from dev branch in dev env, read from master branch in prd env)
Other teams use tagging for versioning notebooks.
We don't need storage for notebooks.(We put output notebooks in gcs)

How do you think ?

onevirus:
- Add GithubHandler
- Add requirements for GithubHandler

Now we can use `https://github.com/.../blob/...ipynb` as input path.
Use `GITHUB_ACCESS_TOKEN` environment variable for accessing private repo.

Fixes #556


MSeal (approved):
Tested it out and it appears to work as advertised. It'd also be neat to support the ssh git uri patterns down the road too.

MSeal:
Error is unrelated to PR -- GCS filesystem had a breaking change it looks like.

PR #624, opened 2021-08-02
Support new version of gcsfs

### LINKED ISSUE #623: Need fix for recent GCFS changes
## Build Issue

GCS dependencies moving forward need a patch for papermill to be able to use the new version of the library.

See: https://github.com/nteract/papermill/actions/runs/1069848310

onevirus:
## What does this PR do?

- Import retry module from gcsfs package

@MSeal 
I tested only latest version of gcsfs. IMHO, it's hard and will be harder to support old gcsfs.
So, how about update minimum version of gcsfs to 'gcsfs>=2021.07.0' and remove old import statements ?

Fixes #623


MSeal (approved):
TY! I really appreciate this fix for the builds.

PR #629, opened 2021-08-30
Bugfix/pickle errors

### LINKED ISSUE #628: PapermillExecutionError cannot be loaded from picklefile
## 🐛 Bug

As part of my team's workflow, we extensively use Papermill to run notebooks as part of our Apache Airflow pipelines. When upgrading Airflow to version 2.X.X we found that our default error handling tools were no longer providing detailed errors from our PapermIll tasks. The cause for this is that Airflow changed the way it handles errors in it's tasks. It temporarily pickles the error message, and then at a later point loads the pickled exception. 

Because PapermillExecutionError is called with multiple arguments, but then only passes up a message argument to its superclass, pickle is unable to load it. See [this stack overflow question/answer](https://stackoverflow.com/questions/41808912/cannot-unpickle-exception-subclass) for a general case of this issue.

This error can be recreated with this simple test:
```{python}
import tempfile
import pickle

from papermill.exceptions import PapermillExecutionError

def test_exceptions_are_unpickleable():
    """Ensure exceptions can be unpickled"""
    temp_file = NamedTemporaryFile()
    err = PapermillExecutionError(1,2, "TestSource", "Exception", Exception(), ["Traceback", "message"])

    with open(temp_file.name, 'wb') as fd:
        pickle.dump(err, fd)
    # Read the Pickled File
    temp_file.seek(0)
    data = temp_file.read()
    pickled_err = pickle.loads(data)
    assert str(pickled_err) == str(err)
```

Calder-Ty:
## What does this PR do?
Adds the ability to load PapermillExecutionErrors from pickle files. 

Airflow 2.0.0 changed the way it handles error reporting by temporarily pickling the exception. Prior to this commit, papermill exceptions could be pickled by airflow, but then not loaded back, causing cryptic error messages. 

This pull request addresses this issue, by modifying the call to the superclass, and sending up all the arguments used to create the exception. It also implements a `__str__` method, to maintain compatibility with the behavior of calling `str()` on the exception returning the formatted message.

Fixes #628

### DIFF
--- a/papermill/exceptions.py
+++ b/papermill/exceptions.py
@@ -23,18 +23,26 @@ class PapermillExecutionError(PapermillException):
     """Raised when an exception is encountered in a notebook."""
 
     def __init__(self, cell_index, exec_count, source, ename, evalue, traceback):
+        args = cell_index, exec_count, source, ename, evalue, traceback
         self.cell_index = cell_index
         self.exec_count = exec_count
         self.source = source
         self.ename = ename
         self.evalue = evalue
         self.traceback = traceback
-        message = "\n" + 75 * "-" + "\n"
-        message += 'Exception encountered at "In [%s]":\n' % str(exec_count)
-        message += strip_color("\n".join(traceback))
-        message += "\n"
 
-        super(PapermillExecutionError, self).__init__(message)
+        super(PapermillExecutionError, self).__init__(*args)
+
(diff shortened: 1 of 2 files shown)


MSeal (approved):

MSeal:
Thanks for the improvement @Calder-Ty !

PR #643, opened 2021-12-07
fix(iorw.py): misleading output message

ksyx:
## What does this PR do?

`papermill foobar.ipynb foobar.ipynb.out` gives me the followings:

```
papermill/iorw.py:126: UserWarning: The specified input file (foobar.ipynb.out) does not end in one of ['.ipynb', '.json']
  "The specified input file ({}) does not end in one of {}".format(path, extensions)
```

Which led me think it took the output path as input file.


willingc (approved):
Thanks for the clarification.

PR #666, opened 2022-05-20
Drop support for Python 3.6

### LINKED ISSUE #665: Python 3.6 EOL
## 🐛 Bug

Python 3.6 reached EOL so it will not receive any more updates, including security updates.
Support for the next version and beyond is considered unnecessary.

(Of course, I know it is not a bug, but it is different from a feature request, so we are creating an issue as a bug.)

reoono:
## What does this PR do?

Updated documentation, setup.py, and CI to drop support for 3.6.

Fixes #665


MSeal (approved):
Thanks!

PR #668, opened 2022-06-16
use unittest.mock instead of mock

### LINKED ISSUE #659: use unittest.mock instead of mock
## 🐛 Bug

Could you please consider unittest.mock instead of mock everywhere consistently or at least use unittest.mock as a fallback when mock is not installed?

reoono:
## What does this PR do?

Changed to use `unittest.mock` instead of `mock`.
- Supported versions of Python have `unittest` as a standard library,
- Unnecessary dependencies can be removed.

Fixes #659

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -4,6 +4,5 @@
 Sphinx==3.5.4
 furo==2021.4.11b34
 myst-parser==0.13.7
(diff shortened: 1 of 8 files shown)


MSeal (approved):
Thanks! This was a left-over from supporting older Python versions

PR #669, opened 2022-06-27
Option for not writing an output ipynb file

dcnadler:
## What does this PR do?

In the current setup of papermill, an `output_path` is required to write an output ipynb file even if you only want to use the returned `NotebookNode` object. For example, in my case, I end up using `tempfile` a good amount to get around this.

This motivation came up in PR https://github.com/nteract/papermill/pull/107, as well

#### Approach

Let user pass `output_path = None` to indicate they don't want to write an output ipynb.

A new IO handler was added, `NoIOHandler` to deal with this option.

#### Alternative approach

I tried to match the api interface of PR https://github.com/nteract/papermill/pull/107, but I could also see having a specific string that indicates no output writing (similar to `"-"` indicating stdin/stdout).

The advantage being that the IO handler can be registered in the IO factory as designed, and not implemented outside of it as I had to do with `path=None`, since [`get_handler`](https://github.com/nteract/papermill/blob/main/papermill/iorw.py#L160) uses `path.startswith`.

If that sounds like a better approach, I'm open to string suggestions - haven't thought of a good one yet :)

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -36,8 +36,8 @@ def execute_notebook(
     ----------
     input_path : str or Path
         Path to input notebook
-    output_path : str or Path
-        Path to save executed notebook
+    output_path : str or Path or None
+        Path to save executed notebook. If None, no file will be saved
     parameters : dict, optional
         Arbitrary keyword arguments to pass to the notebook parameters
     engine_name : str, optional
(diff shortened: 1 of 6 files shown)


MSeal (approved):
Thanks, this seems like a cleaner way than doing '/dev/null' output pathing.

mnp:
How does one do this from the CLI?

Seems like passing /dev/null would be fine but you get a warning when it doesn't know that file extension.

dcnadler:
This PR did modify the CLI interface.

If you wanted to add a no-output CLI option, it looks like you'd need to add that in https://github.com/nteract/papermill/blob/main/papermill/ and work around the stdout logic.

PR #673, opened 2022-07-02
Remove deepcopy steps during parameterize_notebook.

surdouski:
This PR removes the `nb = copy.deepcopy(nb)` steps inside parameterize.py and execute.py. This does not fix any tracked issue I know of. What this PR solves are some issues and annoyances surrounding the deepcopy (and pickle) standard libraries.

```python
# Copy the nb object to avoid polluting the input
nb = copy.deepcopy(nb)
```

Currently, notebooks that contain cells or parameters with objects without a \_\_dict\_\_ require some heavy object/class patching of their magic methods. Also, anything that is unable to be serialized with pickle will run into issues here as well. Removing all of the `copy.deepcopy(nb)` calls fixes this issue.

The removal of these can be done safely because the deepcopy currently has no useful behavior; just looking at the code surrounding these calls makes it somewhat apparent. All tests are currently passing on my machine with the removal of these calls to deepcopy, so there would have to be some undocumented use-cases with no matching tests if there are any issues here.

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 
-import copy
 import nbformat
 from pathlib import Path
 
@@ -148,9 +147,6 @@ def prepare_notebook_metadata(nb, input_path, output_path, report_mode=False):
     report_mode : bool, optional
        Flag to set report mode
     """
-    # Copy the nb object to avoid polluting the input
-    nb = copy.deepcopy(nb)
-
     # Hide input if report-mode is set to True.
     if report_mode:
         for cell in nb.cells:
@@ -185,7 +181,6 @@ ERROR_ANCHOR_MSG = (
 
 
 def remove_error_markers(nb):
-    nb = copy.deepcopy(nb)
     nb.cells = [cell for cell in nb.cells if ERROR_MARKER_TAG not in cell.metadata.get("tags", [])]
     return nb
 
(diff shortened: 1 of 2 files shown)


rohitsanj (approved):
Thanks!

edublancas:
> Currently, notebooks that contain cells or parameters with objects without a __dict__ require some heavy object/class patching of their magic methods. Also, anything that is unable to be serialized with pickle will run into issues here as well. Removing all of the copy.deepcopy(nb) calls fixes this issue.

Hi @surdouski , I'm interested in learning more about this use case; under which conditions a notebook contains non-serializable parameters?

surdouski:
@edublancas Since I won't be updating papermill until there is a new release on pip (or until I decide to build a wheel myself, but I am lazy), you can take a look at [this](https://github.com/ComposableAnalytics/Composapy/tree/main/composapy/patch) directory of my repo where I implement any patching needed. Some of it is for `copy.deepcopy` related things and some of it is `json.loads/dumps` related things. Since I've gotten this taken care of in papermill, the next PR's (when I get around to it) will probably be on nbformat -- related to the usage of json (I might get around to this, I might not, idk).

The reason that I need to patch things is because pythonnet doesn't set attributes on the objects, which means the call to \_\_dict\_\_ raises an exception. In addition, I require custom serialization/deserialization patterns for certain .NET objects such as things containing a Uri.

If you have any specific questions, let me know.

PR #674, opened 2022-07-03
Implement bash translator

kdm9:
## Support translation of parameters to bash/*NIX shell

Thanks for papermill! I have used it for python notebooks, but a lot of my work is best done in shell, for which I use jupyter notebooks with the [bash_kernel](https://github.com/takluyver/bash_kernel/) (via *.sh scripts and jupytext). 

Alas papermill doesn't seem to support bash/shell script, due to a hitherto unimplemented bash translator. This PR implements a `BashTranslator`, and therefore enables bash kernel notebooks to be parametrised.  This code supports `str`, `list`, `bool`[^1], `None`, `int`, and `float` parameter values. Dict could in theory be supported via [associative arrays](https://linuxhint.com/associative_arrays_bash_examples/) but for now they are not implemented[^2].

For a functioning example, please see the attached notebook.

Best,
K

[^1]: Bash doesn't really have a bool type, but `true` and `false` are at least common and sensible string constants for truthy-ness and falsy-ness. See for example [this SO answer](https://stackoverflow.com/a/21210966).
[^2]: Associative arrays require quite recent bash versions, and I've never seen them used outside tutorials. Supporting them would also likely require a more involved code path as one must pre-declare an associative array in bash, and then declare one entry per line (see link above).

### DIFF
--- a/papermill/tests/test_translators.py
+++ b/papermill/tests/test_translators.py
@@ -601,3 +601,49 @@ def test_translator_must_implement_comment():
 
     with pytest.raises(NotImplementedError):
         MyNewTranslator.comment("foo")
+
+
+# Bash/sh section
+@pytest.mark.parametrize(
+    "test_input,expected",
+    [
+        ("foo", "foo"),
+        ("foo space", "'foo space'"),
+        ("foo's apostrophe", "'foo'\"'\"'s apostrophe'"),
+        ("shell ( is ) <dumb>", "'shell ( is ) <dumb>'"),
+        (12345, '12345'),
+        (-54321, '-54321'),
+        (1.2345, '1.2345'),
+        (-5432.1, '-5432.1'),
+        (True, 'true'),
+        (False, 'false'),
+        (None, ''),
+    ],
+)
+def test_translate_type_sh(test_input, expected):
+    assert translators.BashTranslator.translate(test_input) == expected
+
+
+@pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


kdm9:
Alas GH doesn't support attaching *.ipynb, so I have had to rename it *.txt, but here's the test notebook.

[test.ipynb.txt](https://github.com/nteract/papermill/files/9035277/test.ipynb.txt)

The following params.yml can be used with this:

```
say: "Write a book"
people:
  - Alice
  - Bob
  - Robert'); DROP TABLE students;-- # XKCD's favourite son
```

kdm9:
A friendly ping to the nteract team about this PR. All tests pass locally, and this is currently blocking us from using papermill in our analyses.

rohitsanj (changes requested):
Thanks for the PR, @kdm9! Can you please add some tests to [`test_translators.py`](https://github.com/nteract/papermill/blob/main/papermill/tests/test_translators.py).

kdm9:
@rohitsanj tests implemented and passing. There are some unrelated failures on my machine, all to do with the google cloud interaction, which I assume is because I don't have a GCS environment correctly configured. Any chance you can allow the CI to run on this PR, so we can see if all is green in a well-configured environment?

MSeal (approved):
Thanks for adding tests! They passed on the CI chain so no worries about local dev issues there

PR #677, opened 2022-08-26
add dependabot for GHA

Borda:
## What does this PR do?

This will help to keep the CI/CD updated.
Dependable is a GitHub application that, based on config checks if some GitHub action can be updated to the latest version if yes, it will create a PR with this update some it will also verify that the update still works with the repository...

See for example here: https://github.com/Borda/docker_sample/pulls?q=is%3Apr+is%3Aclosed+author%3Aapp%2Fdependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+# Basic dependabot.yml file with minimum configuration for two package managers
+
+version: 2
+updates:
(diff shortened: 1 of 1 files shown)


MSeal (approved):
Ty for adding this! I used this elsewhere and it's a nice addition to repos.

PR #678, opened 2022-08-26
add pre-commit

Borda:
## What does this PR do?

This will automatically format the before each commit if you enable it in your local dev environment.
This can also be paired with Bot (need to be installed), which would commit formatting correction directly to each open PR.

See: https://pre-commit.ci


MSeal (approved):
There was pushback on having pre commit hooks back in the day, but I always wanted to give it a try for papermill. Let's merge and see if folks complain or want it adjusted

Issue #680, opened 2022-09-09
Ansiwrap raises no module found error due to use of deprecated imp command


soggyomelette:
## 🐛 Bug

When bundled using PyInstaller, ansiwrap is unable to find textwrap3 using the imp command leading to a no module found error as shown below.
![image](https://user-images.githubusercontent.com/85685255/189323260-a4dbd229-84b6-4499-bcdb-634a96c94419.png)

Since the repo hasn't been updated in ages (https://github.com/jonathaneunice/ansiwrap) and there is only one instance of its usage in papermill, the best solution should be to remove its usage from the repo.

MSeal:
This could be resolve with the merged PR and in the next release.

piiq:
Hi @MSeal, thanks for reviewing and merging https://github.com/nteract/papermill/pull/681
Is there a timeline for a release that would include this change and is there anything I can do to help making that release?

PR #681, opened 2022-09-09
Remove use of ansiwrap

soggyomelette:
## What does this PR do?

Removes use of ansiwrap in the package for the reasons in #680

### DIFF
--- a/papermill/exceptions.py
+++ b/papermill/exceptions.py
@@ -1,8 +1,5 @@
 # -*- coding: utf-8 -*-
 
-from ansiwrap import strip_color
-
-
 class AwsError(Exception):
     """Raised when an AWS Exception is encountered."""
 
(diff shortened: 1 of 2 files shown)


MSeal (approved):
Thanks for identifying and opening the PR for changes!

dhirschfeld:
It would be great to get this fix in a release as, now that Python 3.12 is released, this is causing issues for any packages depending on `papermill`.

MSeal:
Got it, alright I can prepare a release of papermill to account for recent python changes. There's been a backlog of issues / PRs that will need to be a put into a fixed release build.

dhirschfeld:
On the `conda-forge` front, we've patched the repodata so there's no longer breakage, but instead you  won't be able to install `papermill` in a 3.12 environment... at least until the new version is published.

MSeal:
Working through some other packaging issues, but will have a 3.12 build out once those are solved

MSeal:
@dhirschfeld 2.5.0 is released to PyPI now

PR #693, opened 2022-10-04
apply pre-commit

Borda:
## What does this PR do?

It is just running `pre-commit run --all-files` but it would be better to install a bot that would be fixing this automatically for each PR - https://pre-commit.ci/

Follow-up of #678


MSeal (approved):
ty!

PR #694, opened 2022-10-05
Remove deepcopy

surdouski:
Just found this one after updating something internal and hitting serialization issue again. See previous PR regarding deepcopy (they have no purpose).


MSeal:
There is a purpose to deepcopy, in the interface, but I agree removing it should be expected at this point. Very early on there was more manipulation of nbformat objects and some implicit expectations that the input wouldn't be modified. Mostly when scrapbook code was still part of papermill. Not as much of a concern now-a-days so happy to merge it.

MSeal (approved):

surdouski:
I figured I could quickly update this in browser, I'll wait till I'm home later today to actually pull everything down and run the tests locally before committing anything else to the branch.

surdouski:
@MSeal Can this be merged? Thanks.

MSeal (approved):

PR #728, opened 2023-08-01
ABS add support for using a Service principle via EnvCreds

NickSchouten:
It allows abs:// to be used without a sas token. In this case it will fall back on[ Azure identities Environment credentials](https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.environmentcredential?view=azure-python). This allows you to use a service principle for authentication with azure.

Changes:
abs.py changed the regex
abs.py added a default for when sas token is empty. It will use env creds in this case. 
tests for ABS without a sas token

Dependecies:
azure-identity -- because we will be using EnvironmentCredential from it

### DIFF
--- a/papermill/abs.py
+++ b/papermill/abs.py
@@ -3,6 +3,7 @@ import re
 import io
 
 from azure.storage.blob import BlobServiceClient
+from azure.identity import EnvironmentCredential
 
 
 class AzureBlobStore(object):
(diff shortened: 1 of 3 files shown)


MSeal (approved):

MSeal:
Thanks for the addition! Will get it into this next release.

PR #738, opened 2023-11-09
Implement --remove-tagged-cells

astrofrog:
This implements a new --remote-tagged-cells option as described by @MSeal in https://github.com/nteract/papermill/issues/429#issuecomment-642806948. The idea of the current implementation is to keep it as simple as possible – only being able to specify a single tag and requiring an exact match (no regular expressions etc). My personal use case for this is being able to run notebooks with/without diagnostic plotting but there are other use cases described in https://github.com/nteract/papermill/issues/429.

If the maintainers are on board with this PR so far, I can add documentation and anything else required.

One thought I had was that perhaps this option would read more clearly as ``--remove-cells-with-tag`` otherwise it sounds like a boolean option, and also doesn't make it clear that a single tag is what should be passed in. What do people think about this?

### DIFF
--- a/papermill/cli.py
+++ b/papermill/cli.py
@@ -63,6 +63,9 @@ def print_papermill_version(ctx, param, value):
     '--parameters_base64', '-b', multiple=True, help='Base64 encoded YAML string as parameters.'
 )
 @click.option(
+    '--remove-tagged-cells', type=str, help='Remove cells with the specified tag before execution.'
+)
+@click.option(
     '--inject-input-path',
     is_flag=True,
     default=False,
@@ -166,6 +169,7 @@ def papermill(
     parameters_file,
     parameters_yaml,
     parameters_base64,
+    remove_tagged_cells,
     inject_input_path,
     inject_output_path,
     inject_paths,
@@ -259,6 +263,7 @@ def papermill(
             request_save_on_cell_execute=request_save_on_cell_execute,
             autosave_cell_every=autosave_cell_every,
             prepare_only=prepare_only,
+            remove_tagged_cells=remove_tagged_cells,
             kernel_name=kernel,
             language=language,
             progress_bar=progress_bar,
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -2,6 +2,7 @@
 
 import nbformat
 from pathlib import Path
+from copy import deepcopy
 
 from .log import logger
 from .exceptions import PapermillExecutionError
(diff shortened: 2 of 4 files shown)


willingc:
Thanks @astrofrog (nice to see you here). It looks like some of the CLI tests still need updating. It also looks like botocore is being grumpy with python 3.12.

MSeal:
Thanks @astrofrog -- on the botocore build part I am working on getting a new release out anyway and need to fix dep updates since the last one regardless.

MSeal:
Looks like tests are failing now due to some test scaffolding including the input_path in the arguments.

papermill/cli.py
```diff
@@ -61,6 +61,9 @@ def print_papermill_version(ctx, param, value):
 @click.option(
     '--parameters_base64', '-b', multiple=True, help='Base64 encoded YAML string as parameters.'
 )
+@click.option(
+    '--remove-tagged-cells', type=str, help='Remove cells with the specified tag before execution.'
```
akx, line 65: What about
```suggestion
    '--remove-cells-tagged', type=str, help='Remove cells with the specified tag before execution.'
```

Then it'd be pretty fluent on the command line:

```
--remove-cells-tagged=a-tag
```

papermill/execute.py
```diff
@@ -104,6 +109,8 @@ def execute_notebook(
             )
 
         nb = prepare_notebook_metadata(nb, input_path, output_path, report_mode)
+        if remove_tagged_cells is not None:
```
akx, line 112:
```suggestion
        if remove_tagged_cells:
```
should be enough – the empty string isn't a valid tag, right..?

papermill/execute.py
```diff
@@ -165,6 +172,33 @@ def prepare_notebook_metadata(nb, input_path, output_path, report_mode=False):
...
+    Parameters
+    ----------
+    nb : NotebookNode
+       Executable notebook object
+    tag : str
+        Tag to used to identify cells to remove.
+    """
+
+    # Copy the notebook to avoid changing the input one
+    nb = deepcopy(nb)
```
akx, line 188: E.g. `remove_error_markers` modifies the notebook in place...

andreytaboola:
Any updates on this? I am also interested in this feature and will appreciate push on it

wd60622:
What is needed to get this over the finish line? Would love to use this feature and willing to help out if possible 😀

willingc:
Ping @MSeal. Do you have time to reboot this? Thanks.

atmorling:
If it's helpful - I have [this](https://github.com/atmorling/papermill/tree/am/remove-tagged-cells) branch which builds off of this PR with @akx's feedback merged + fixing the failing tests.
Happy to raise a seperate PR on this repo or to the branch from @astrofrog's fork - whatever is easier for pushing this through.

PR #739, opened 2023-11-13
move `black` & `flake8` to precommit

Borda:
## What does this PR do?

It shall simplify maintenance since we already have the preco-commit configuration with installing this [precommit bot](https://github.com/marketplace/pre-commit-ci) (free for public repos); it will eventually apply fixes for incorrect formatting so the user doe/contributor does not need to think about it... I have been using it on several other projects and it is very helpful without any overhead... :rabbit: 

Also, after adding the bot requited check will need to be updated


.github/workflows/ci.yml
```diff
@@ -29,8 +29,8 @@ jobs:
           set -xe
           python -VV
           python -m site
-          python -m pip install --upgrade pip setuptools wheel
-          python -m pip install --upgrade coverage[toml] virtualenv tox tox-gh-actions
+          python -m pip install --upgrade \
```
willingc, line 32: Let's make this all one line since we are not picky about line length and it improves readability.

willingc (commented):
Looks good to me. Thanks @Borda for improving this. One small change is requested and then I'm happy to approve.

.github/workflows/ci.yml
Borda: Shrinked to one line

MSeal (approved):
Thanks for making the requested change! I'll let @willingc do the merge in case she wants anything else adjusted :)

willingc (approved):
Thanks for the fix and the improvements. Happy to merge. 👍🏼

Borda:
> Thanks for the fix and the improvements. Happy to merge. 👍🏼

Could you pls install the precommit bot and set it as required check so we preserve the code quality bar... 🐰 

Probably it need to be someone with org admin 🦩

willingc:
@Borda I think that I have precommit setup now. Please check to see if it looks correct to you on https://github.com/nteract/papermill/pull/738

Borda:
> @Borda I think that I have precommit setup now. Please check to see if it looks correct to you on #738

I don't see it there, except the config file the bot need to be installed

PR #740, opened 2023-11-14
fix: Don't pass `input_path` to `PapermillNotebookClient.__init__`

kdestin:
## What does this PR do?

This pull request is a minor fix that prevents NBClientEngine from sending the `input_path` parameter to `PapermillNotebookClient`, which appears to not actually accept a parameter of that name.

# Background 

This change resolves a deprecation warning raised by `traitlets` [when a trait provides an unrecognized argument to `super().__init__`](https://github.com/ipython/traitlets/blob/008708a87d320e6e237ae370d60037889fd22e41/traitlets/traitlets.py#L1387-L1398):

```
  /home/user/path/to/venv/lib/python3.9/site-packages/traitlets/traitlets.py:1387: DeprecationWarning: Passing unrecognized arguments to super(PapermillNotebookClient).__init__(input_path='/path/to/notebook.ipynb').
  object.__init__() takes exactly one argument (the instance to initialize)
  This is deprecated in traitlets 4.2.This error will be raised in a future release of traitlets.
    warn(
```

The warning appears to be caused by `papermill_engines.execute_notebook_with_engine`

https://github.com/nteract/papermill/blob/47c4af532ec26b7f73be7834350b8a5126deaa11/papermill/execute.py#L119-L131

which forwards an `input_path` parameter to `NBClientEngine.execute_managed_notebook`, which is then supplied to `PapermillNotebookClient.__init__`. Neither `PapermillNotebookClient` nor its super classes appear to accept an `input_path` parameter, so the remedy appeared to be to remove it from the provided kwargs. 

## Motivation

Running `papermill` in a `pytest` test without `pytest` reporting the surfaced warnings.

---

Please let me know if this actually needs test coverage to ensure that a deprecationwarning isn't being raised.


MSeal (approved):
Thanks for catching this. That warning was annoying to see. No need for a test for this, though pro-tip for future PRs, keep the linting changes across the repo and the actual code change as separate PRs. It makes it difficult to review, revert, and almost immediately causes conflicts that couple cleanup code with implementation code inseparably.

PR #751, opened 2023-11-21
add dependabot also for pip packages

Borda:
## What does this PR do?

Recently, I got some vulnerability warnings.


willingc (approved):
Thanks @Borda

PR #752, opened 2023-11-21
switch flake8 -> ruff

Borda:
## What does this PR do?

Ruff has been adopted by several major projects already I would consider it as stable and offer significant performance boost and fixing multiple lints in the same time


willingc (approved):
I ❤️ Ruff. Thanks for this PR @Borda. Very happy to merge.

PR #755, opened 2023-11-21
update Black with line 120

Borda:
## What does this PR do?

seems that the black config was a dummy one with lots of not-relevant settings to this project

Moreover changing the line length to 120 as it is already with flake8 and it common standard now


.pre-commit-config.yaml
```diff
@@ -44,7 +44,6 @@ repos:
     hooks:
       - id: black
         name: Black code
-        args: ["-S"]
```
Borda, line 47: This has been already in the config as `skip-string-normalization = true`

willingc:
@Borda If we use Ruff, I believe that we can remove black.

Borda:
> @Borda If we use Ruff, I believe that we can remove black.

tbh, never tried but still, I may tend to keep it... of have the same setting? :)

willingc:
Let's do this @Borda. I'll go ahead and merge this since I am a fan of 120.

FYI on the differences between Ruff and Black for formatting. They are fairly subtle at the moment. https://docs.astral.sh/ruff/formatter/#the-ruff-formatter

willingc (approved):

PR #756, opened 2023-11-21
bump syntax helper to 3.8

Borda:
## What does this PR do?

the `setup.py` has already set the min python version as 3.8 so we shall keep this aligned


willingc (approved):
Good call.

PR #763, opened 2023-12-18
Pre-commit fixes

akx:
## What does this PR do?

This PR cleans up the pre-commit configuration slightly:

* Python 3.8 is no longer required to run the tools
* yesqa isn't necessary, since it's depending on flake8 (the use of which was dropped in #752)
* Ruff has rules for Pyupgrade, so it isn't necessary as a separate hook either

Some of the f-string changes the pyupgrade commit does are also done in #762, so these will conflict slightly.

I didn't enable `ruff-format`, since the current quote style in this repository is mixed, and that's not (yet: https://github.com/astral-sh/ruff/issues/9185) supported by Ruff's formatter.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,3 @@
-default_language_version:
-  python: python3.8
-
 ci:
   autofix_prs: true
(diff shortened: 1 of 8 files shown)


MSeal (approved):

MSeal:
Thanks for the cleanup!

PR #778, opened 2024-02-06
make progrgress_bar param accept a dict

### LINKED ISSUE #777: Make the parameters of progress_bar specifiable by a dictionary
Dear developers, thank you for maintaining such a wonderful tool.

## 🚀 Feature
I want to make the parameters of progress_bar specifiable by a dictionary.

### Motivation
When executing a large number of notebooks using `papermill.execute.execute_notebook`, I want to set `leave=False` to limit the progress bar displayed by tqdm.

It seems that `papermill.execute.execute_notebook` calls `NotebookExecutionManager` by default, so if we make the progress_bar parameter specifiable as a dict, it might be possible to pass parameters to tqdm.

https://github.com/nteract/papermill/blob/5addc11f2aa36d5ed18d21fb53f4680b62cf8e3c/papermill/engines.py#L108-L112

wsuzume:
Fixes #777 (Oh, I unintendedly hit the jackpot🎰)

## What does this PR do?
This pull request is an extension that allows passing arguments to `tqdm` by enabling the `progress_bar` parameter of `execute_notebook` to accept a `dict`.

This pull request maintains compatibility with the current implementation, though it might make the conditional branching a bit messy...

I believe there could be discussions on how to actually enable such specification, as follows.

```python
            # Ensure at least unit and desc.
            # Conversely, it becomes impossible to remove unit and desc.
            elif isinstance(progress_bar, dict):
                _progress_bar = { "unit": "cell", "desc": "Executing" }
                _progress_bar.update(progress_bar)
                self.pbar = tqdm(total=len(self.nb.cells), **_progress_bar)
```

```python
            # Simple and accepts any dictionary.
            # However, it's not intuitive as an empty dictionary is precluded by `if progress_bar:`.
            elif isinstance(progress_bar, dict):
                self.pbar = tqdm(total=len(self.nb.cells), **progress_bar)
```

```python
        # It becomes possible to accept an empty dictionary,
        # but the implementation becomes slightly complicated.
        if isinstance(progress_bar, bool):
            if progress_bar:
                # lazy import due to implicit slow ipython import
                from tqdm.auto import tqdm
                ...
        elif isinstance(progress_bar, dict):
            # lazy import due to implicit slow ipython import
            from tqdm.auto import tqdm
            ...
```

### DIFF
--- a/papermill/engines.py
+++ b/papermill/engines.py
@@ -109,7 +109,14 @@ class NotebookExecutionManager:
             # lazy import due to implicit slow ipython import
             from tqdm.auto import tqdm
 
-            self.pbar = tqdm(total=len(self.nb.cells), unit="cell", desc="Executing")
+            if isinstance(progress_bar, bool):
+                self.pbar = tqdm(total=len(self.nb.cells), unit="cell", desc="Executing")
+            elif isinstance(progress_bar, dict):
+                _progress_bar = { "unit": "cell", "desc": "Executing" }
+                _progress_bar.update(progress_bar)
+                self.pbar = tqdm(total=len(self.nb.cells), **_progress_bar)
+            else:
+                raise TypeError(f"progress_bar must be instance of bool or dict, but actual type '{type(progress_bar)}'.")
 
     def now(self):
         """Helper to return current UTC time"""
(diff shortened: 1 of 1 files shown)


MSeal (approved):
This seems like a great improvement for the control plane of progress bars. Thanks for exploring and adding it!

PR #779, opened 2024-02-14
Upgrade tests to moto v5

bnavigator:
Moto v5 [replaced `mock_s3` with `mock_aws`](https://docs.getmoto.org/en/latest/docs/releases.html#breaking-changes)

Supersedes #775

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -6,7 +6,7 @@ google_compute_engine # Need this because boto has issues with dynamic package l
 ipython>=5.0
 ipywidgets
 notebook
-moto ~= 5.0
+moto >= 5.0.0,<5.1.0
 pytest>=4.1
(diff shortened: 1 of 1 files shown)


requirements/dev.txt
```diff
@@ -6,7 +6,7 @@ google_compute_engine # Need this because boto has issues with dynamic package l
 ipython>=5.0
 ipywidgets
 notebook
-moto <5.0.0
+moto ~= 5.0
```
Borda, line 9:
```suggestion
moto >= 5.0.0,<5.1.0
```
lets keep the same notation

Borda (approved):

MSeal:
Failure that's left is already fixed in main -- going to merge instead of trying to pull the base up to the PR

Issue #501, opened 2020-05-11
No error message on magic command error


totycro:
Hi,

When I execute a notebook with papermill and a magic command fails, I don't actually get an error message and also the exit code is 0, even though the notebook execution stopped at this point and `metadata.papermill.exception` is set to `true` in the notebook.

I would expect to see a python exception and a non-zero exit code in this case as the execution actually failed.

I've attached a sample input 
[simple.ipynb.txt](https://github.com/nteract/papermill/files/4610414/simple.ipynb.txt) and output [simple_out.ipynb.txt](https://github.com/nteract/papermill/files/4610413/simple_out.ipynb.txt). This is the output i get with debug logging turned on:
```
$ papermill --log-level DEBUG simple.ipynb simple_out.ipynb
Input Notebook:  simple.ipynb
Input Notebook:  simple.ipynb
Output Notebook: simple_out.ipynb
Executing:   0%|                                                            | 0/2 [00:00<?, ?cell/s]Using selector: EpollSelector
Starting kernel (async): ['/opt/conda/bin/python', '-m', 'ipykernel_launcher', '-f', '/tmp/tmpap6uw67z.json', '--HistoryManager.hist_file=:memory:']
Connecting to: tcp://127.0.0.1:57435
connecting shell channel to tcp://127.0.0.1:49455
Connecting to: tcp://127.0.0.1:49455
connecting iopub channel to tcp://127.0.0.1:33771
Connecting to: tcp://127.0.0.1:33771
connecting stdin channel to tcp://127.0.0.1:44405
Connecting to: tcp://127.0.0.1:44405
connecting heartbeat channel to tcp://127.0.0.1:33255
Using selector: EpollSelector
connecting control channel to tcp://127.0.0.1:57435
Connecting to: tcp://127.0.0.1:57435
Executing notebook with kernel: python3
Executing cell:
%env undefined
msg_type: status
content: {'execution_state': 'busy'}
msg_type: execute_input
content: {'code': '%env undefined', 'execution_count': 1}
msg_type: stream
content: {'name': 'stderr', 'text': 'UsageError: Environment does not have key: undefined\n'}
msg_type: status
content: {'execution_state': 'idle'}
Executing:  50%|██████████████████████████                          | 1/2 [00:00<00:00,  1.29cell/s]
$ echo $?
0
```

I'm using papermill 2.1.0.

MSeal:
[...]

Issue #381, opened 2019-06-20
Does not catch certain query magic error? (UsageError: Cell magic `%%bigquery` not found.)


lucidsushi:
Normally without running `%load_ext google.cloud.bigquery` , using `%%bigquery` would throw this error, but when executing a notebook via  papermill this doesn't seem to be caught  as an exception

MSeal:
Likely this is a result of the magic function not raising an exception. I'd need to investigate some to determine if that's the case. If so, then this should probably be an issue on that project to raise rather than silently fail.

lucidsushi:
If it's any help at all, I traced the error back to ipython (https://github.com/ipython/ipython/blob/a8165da9102a62203c824f8cb59988e188fc6032/IPython/core/magics/basic.py), where it raised the `UsageError`, which indeed you see it in the output of cells in the notebook, but it's not acting like an exception

MSeal:
Sorry I got very sick and backlogged on items.

If it's raising `UsageError` from the linked code it should halt and send an error status that halts execution as you'd expect. Do you have a minimal working example I could reproduce with? That'd help with debugging at this point.

VasuBhog:
@MSeal I am also seeing this error occur with kql magic command (%%kql) which executes the parameterized notebook but then has an error later on the usage.

```
user % papermill simpleKQLmagic3.ipynb output_simple.ipynb
Input Notebook:  simpleKQLmagic3.ipynb
Output Notebook: output_simple.ipynb
Executing:   0%|                                                      | 0/7 [00:00<?, ?cell/s]Executing notebook with kernel: python3
Executing:  43%|███████████████████▋                          | 3/7 [00:08<00:10,  2.60s/cell]2020-11-25 16:29:03,346 - INFO - start parent_monitor for parent id: 20462
2020-11-25 16:29:03,346 - INFO -  * parent id: 20462
2020-11-25 16:29:03,346 - INFO -  * Base folder: /Users/xxx
2020-11-25 16:29:03,346 - INFO -  * Folder list: ['.kqlmagic/temp_files/b40af28a-42a4-4b47-8188-953c3f57d3cc']
2020-11-25 16:29:03,346 - INFO -  * 
2020-11-25 16:29:03,346 - INFO -  **** 
 * Serving Flask app "kqlmagic_temp_files_server" (lazy loading)
 * Environment: development
 * Debug mode: off
 * Running on http://127.0.0.1:50197/ (Press CTRL+C to quit)
[...]

PR #786, opened 2024-03-13
raise PapermillExecutionError when CellExecutionError is raised without cell error output

mbektas:
When notebooks with cell / line magics are executed and if the magic command fails, papermill doesn't throw any exception and returns exit code 0. This is because papermill relies on cell error output to raise PapermillExecutionError.

When magic commands fail, CellExecutionError is still thrown and recorded as cell metadata, and notebook execution stops. However, since these errors don't produce any cell error output, no exception is raised.

Changes in this PR check cell metadata for exceptions recorded by papermill (due to CellExecutionError) in addition to cell error outputs and raise PapermillExecutionError in those cases.

fixes https://github.com/nteract/papermill/issues/501, https://github.com/nteract/papermill/issues/381

### DIFF
--- a/papermill/execute.py
+++ b/papermill/execute.py
@@ -196,22 +196,33 @@ def raise_for_execution_errors(nb, output_path):
     """
     error = None
     for index, cell in enumerate(nb.cells):
-        if cell.get("outputs") is None:
-            continue
+        # check if there is any cell error output
+        if "outputs" in cell:
+            for output in cell.outputs:
+                if output.output_type == "error":
+                    if output.ename == "SystemExit" and (output.evalue == "" or output.evalue == "0"):
+                        continue
+                    error = PapermillExecutionError(
+                        cell_index=index,
+                        exec_count=cell.execution_count,
+                        source=cell.source,
+                        ename=output.ename,
+                        evalue=output.evalue,
+                        traceback=output.traceback,
+                    )
+                    break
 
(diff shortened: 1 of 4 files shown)


ibdafna:
@MSeal apologies for the direct ping. Let us know if there's anything we can do to get this one over the line. Many thanks

MSeal (approved):
There's a subtle case where non-blocking exceptions are raised are intended to not fail cell execution, but it's a much smaller edge case than the exception with no output this PR is fixing (plus I believe other interfaces are leaning into the same tradeoff). Thanks for putting this together!

mbektas:
thanks for the review @MSeal! I ran unit tests locally and they were passing. I will look into failing CI checks.

MSeal:
@mbektas No problem. I should be checking these more regularly.

The current test failure is unrelated to this PR. If you have a few minutes to add a test for the new exception detection it'd be appreciated. Independently I will go fix the unrelated test if you don't patch it yourself and get this merged when I get back from this work trip.

mbektas:
> The current test failure is unrelated to this PR.
I fixed the failing test, it was due to a deprecated pytest feature. I will also be adding test(s) for the magic command error handling.

mbektas:
@MSeal I added a unit test for magic command failures as well. Looks like workflow needs your approval to run again.

PR #788, opened 2024-04-02
prevent error override, fix traceback type

mbektas:
this is in addition to https://github.com/nteract/papermill/pull/786 
- fixing overriding regular error type detection. 
- making traceback an empty list instead of empty string


MSeal:
Thanks!

PR #791, opened 2024-04-03
bring back strip_color and remove ANSI color codes from exception traceback

mbektas:
https://github.com/nteract/papermill/pull/681 removed `ansiwrap` dependency and `strip_color` used to clean up ANSI color codes from exception outputs. That is a breaking change and preventing us to use the latest papermill version.

This PR brings in only the `strip_color` implementation from `ansiwrap` package and uses it to remove ANSI color codes as the versions `<= 2.4.0` did.


mbektas:
@MSeal would you be able to check out this one as well before the next release? thanks!

papermill/exceptions.py
```diff
@@ -75,3 +78,20 @@ def missing_dep():
         )
 
     return missing_dep
+
+
+# copied from https://github.com/jonathaneunice/ansiwrap/blob/master/ansiwrap/core.py
```
MSeal, line 83: Given it's a straight copy there's a bit of ambiguity on if this requires a copy of the license file locally. Specifically it's a single simple regex which only really has one possible implementation in Python, so the condition on if it requires a license copy or is a common-approach change is hard to say perfectly. However, given that code itself is also copied (or expected to be sourced from anscicolor https://github.com/jonathaneunice/ansiwrap/blob/master/ansiwrap/core.py#L35C11-L35C21) I think the authors would consider it the former simple pattern case. In looking at that though perhaps we want to add a dependency on [ansicolor](https://pypi.org/project/ansicolors/) which has no non-dev dependencies for the strip_color function to avoid the copy alltogether? It's probably fine either way though

mbektas: sure. let me use ansicolor package instead.

Issue #765, opened 2024-01-01
Does tqdm print properly with --log-output?


hujc7:
## 🐛 Bug

Not sure if it's a bug or it's expected. When I use tqdm in notebook and use the `--log-output` option, the progress bar keeps printing to a new line. Anyone having the same problem?

Notebook cell
```
from time import sleep
from tqdm import tqdm
 
for i in tqdm(range(int(5)), leave=True):
    sleep(0.1)
```
CMD
```
papermill papermill-test.ipynb papermill-test-out.ipynb --log-output
```
Output
```
Input Notebook:  papermill-test.ipynb
Output Notebook: papermill-test-out.ipynb
Executing notebook with kernel: python3
Executing Cell 1---------------------------------------
  0%|                    | 0/5 [00:00<?, ?it/s]
 20%|██▍         | 1/5 [00:00<00:00,  9.99it/s]
 40%|████▊       | 2/5 [00:00<00:00,  9.88it/s]
 60%|███████▏    | 3/5 [00:00<00:00,  9.85it/s]
 80%|█████████▌  | 4/5 [00:00<00:00,  9.83it/s]
100%|████████████| 5/5 [00:00<00:00,  9.83it/s]
100%|████████████| 5/5 [00:00<00:00,  9.82it/s]


Ending Cell 1------------------------------------------
```

tqdm: 4.65.2
papermill: 2.5.0

Sarius2009:
Same probelm here, would be really nice if this could be fixed!

Issue #496, opened 2020-04-29
log_output argument didn't work for papermill.execute


JustinHu-Drgn:
Hiya, we want to output the cell into logger with papermill, and it was fine for cmd line with papermill directly:
<img width="974" alt="447e52e6-6766-4672-86af-80feb02b37c7" src="https://user-images.githubusercontent.com/49478050/80564152-9cf2d280-8a1f-11ea-8afc-e464d18ddd0c.png">
however it failed to output log with papermill.execute:
![Screenshot 2020-04-29 at 1 47 12 PM](https://user-images.githubusercontent.com/49478050/80564258-f5c26b00-8a1f-11ea-8665-b1af39c0739a.png)

could someone help to debug about it? many thx.

DustinKLo:
@JustinHu-Drgn 
I'm running into the same issue, did you ever figure out a workaround?
thanks 👍

Issue #433, opened 2019-09-19
live printing in notebook


germayneng:
Hi! 

Thanks for creating such a great package. In my notebook, for example if i am training a ML model, typically in the notebook, the training process will output a verbose each step. But if i were to run using papermill, the particular cell will only show the output only after the cell is done executing. Is there any options for CLI to output the verbose message live rather than after the cell is done running? 

e.g 

``` 
training round 1...
training round 2.. 

...etc
```

MSeal:
Yes. You can use `--log-output` and it will stream the logs to the command line live. To note, older versions of nbconvert won't live log, so make sure you have nbconvert >=5.6

SultanOrazbayev:
When I ran a notebook from within a python script using:
```
papermill.execute_notebook(input_path=nbk_path, output_path=nbk_log_path, parameters=myparameters, kernel_name='python3', log_output=True, autosave_cell_every=60)
```
I don't see the output in my shell and also I don't see the print results of a cell that is still running (I was hoping this would be achieved by the autosave every 60 minutes). Is this expected when executing the notebook from within another script/notebook?

MSeal:
So when setting `log_output` in a python script it's making python log the output to the `logging.getLogger('papermill')`. This lets you control how the log message propegates e.g.:

```
logging.basicConfig(level='INFO', format="%(message)s")
```

would setup your default logging experience to print to stderr.

Another option is to set the `stdout_file` and/or `stderr_file` which additional writes if set:

```
result = pm.execute_notebook(
    input_path=nbk_path,
    output_path=nbk_log_path,
    parameters=myparameters,
    kernel_name='python3', 
    log_output=True,
    autosave_cell_every=60,
    stdout_file=sys.stdout,
    stderr_file=sys.stderr,
)
```

As to the autosave, that feature saves the output_path notebook in a regular pattern so you can view the results before a cell is completed, rather than having anything to do with the logging of the process.

SultanOrazbayev:
This is great, thank you!

PR #810, opened 2024-10-18
Fix log output

bryanvaz:
Fixes issue where `\n` characters are randomly inserted into the console for stdout messages. This results in non-deterministic output breaking any downstream process.

Connected to #433, #496, #765 

### Current Behaviour:
When there is execution significant time  or async activity between stdout/stderr calls in an output cell, the default logger inserts newline terminators to the papermill console output. This output is different than output as executed in a live notebook client. (See below for screenshots) 

**Example 1:**
**Notebook Output:**
Screenshot 2024-10-18 at 3 17 11 AM
**Console Output:**
Screenshot 2024-10-18 at 3 19 02 AM

=============

**Example 2:
Notebook Output:**
Screenshot 2024-10-18 at 3 25 31 AM
**Console Output:**
Screenshot 2024-10-18 at 3 24 30 AM

=============

**Example 3:
Notebook Output:**
Screenshot 2024-10-18 at 3 27 04 AM
**Console Output:**
Screenshot 2024-10-18 at 3 27 32 AM

### Expected Behaviour:
When using the `--log-output` option, especially when using the cli, the cell output text from papermill should match the text output when running a notebook in jupyter (or other `.ipynb` client)

### Fix:
Added a secondary logger `notebook_logger` that removes the extraneous terminator from the stream handlers. The notebook logger is only used by cell outputs.

Users can override this behaviour when in using papermill as a library by either resetting the notebook_logger or modifying the parameters when instantiating a Papermill client.

**Output with fix:**
Screenshot 2024-10-18 at 3 47 10 AM

### Reasoning behind Fix
The papermill client was using the default `log.info` of the default notebook client for all stdout messages from a notebook when the explicit `—log-output` option is used. This calls the default logger which is built as a logger for syslog. As a result, the log formatter will automatically add `\n` when flushing. 

In Jupyter, the input cell’s stdout and stderr calls are captured and redirected to the notebook’s output cells without additional `\n` characters added in order to preserve the intended output formatting of the notebook’s author. If the author needs to modify their output to ensure it is compatible with whatever downstream tool/process, they should be responsible for making those changes in the notebook itself and trusting that papermill will not alter the cell’s output.

Unfortunately this does not fix the same issue in the underlying nbclient. As a result, the cell output in the output notebook is still different than if you were to run the notebook in a live system (additional newline characters.) 

Excerpt from python `logging` that is causing the issue:
```python
class StreamHandler(Handler):
…
    terminator = '\n'
…
    def emit(self, record):
        """
        Emit a record.

        If a formatter is specified, it is used to format the record.
        The record is then written to the stream with a trailing newline.  If
        exception information is present, it is formatted using
        traceback.print_exception and appended to the stream.  If the stream
        has an 'encoding' attribute, it is used to determine how to do the
        output to the stream.
        """
        try:
            msg = self.format(record)
            stream = self.stream
            # issue 35046: merged two stream.writes into one.
            stream.write(msg + self.terminator)
            self.flush()

```

### DIFF
--- a/papermill/tests/test_engines.py
+++ b/papermill/tests/test_engines.py
@@ -9,7 +9,7 @@ from nbformat.notebooknode import NotebookNode
 from .. import engines, exceptions
 from ..engines import Engine, NBClientEngine, NotebookExecutionManager
 from ..iorw import load_notebook_node
-from ..log import logger
+from ..log import logger, notebook_logger
 from . import get_notebook_path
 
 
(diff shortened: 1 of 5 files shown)


papermill/tests/test_engines.py
```diff
@@ -434,39 +434,50 @@ def test_nb_convert_engine_execute(self):
...
-                            call('Ending Cell 1------------------------------------------'),
-                            call('Executing Cell 2---------------------------------------'),
-                            call('None\n'),
-                            call('Ending Cell 2------------------------------------------'),
-                        ]
-                    )
-                    warning_mock.is_not_called()
+                with patch.object(notebook_logger, 'info') as notebook_info_mock:
+                    with patch.object(notebook_logger, 'warning') as notebook_warning_mock:
+                        with patch.object(NotebookExecutionManager, 'save'):
```
Borda:
```suggestion
with patch.object(logger, 'info') as info_mock, patch.object(logger, 'warning') as warning_mock, patch.object(notebook_logger, 'info') as notebook_info_mock, patch.object(notebook_logger, 'warning') as notebook_warning_mock, patch.object(NotebookExecutionManager, 'save'):
```
consider reducing the depth

PR #822, opened 2025-08-21
Drop the dependency on ansicolors

musicinmybrain:
## What does this PR do?

The [`ansicolors` package](https://pypi.org/project/ansicolors/) is not actively maintained (last released in 2017), and the function we use is just based on a single simple regular expression. Copy the regular expression into `papermill` and drop the external dependency.


Borda (approved):

willingc (approved):
Thanks for the fix @musicinmybrain. :tada:

PR #823, opened 2025-08-27
chore: Replace deprecated utcnow()

winniehell:
## What does this PR do?

replace `datetime.datetime.utcnow()` with `datetime.datetime.now(datetime.UTC)` to resolve deprecation warnings

see warning in Python documentation: https://docs.python.org/3.9/library/datetime.html#datetime.datetime.utcnow


rgbkrk:
Odd that can't trigger it from here. When I get back to my laptop I'll change the repo settings to modern GitHub actions so we can kick this off and clean up the backlog of PRs.

winniehell:
according to https://docs.github.com/en/actions/how-tos/manage-workflow-runs/approve-runs-from-forks, there should be an approve button here on the conversations tab though I'm not sure where or how it looks like

rgbkrk:
Alright, I think you can now rebase this PR.

PR #846, opened 2025-11-14
Unskip tests that were previously failing

### LINKED ISSUE #831: Fix skipped tests in test_execute
## 🐛 Bug

Two tests were skipped in test_execute.py when getting CI back to green and updating Python support. 

These tests need debugging.

willingc:
Fixes #831 

Tested locally with 3.13 which was failing earlier in the day.

### DIFF
--- a/papermill/tests/test_execute.py
+++ b/papermill/tests/test_execute.py
@@ -174,7 +174,6 @@ class TestBrokenNotebook2(unittest.TestCase):
     def tearDown(self):
         shutil.rmtree(self.test_dir)
 
-    @unittest.skip("FIX: Test is broken")
     def test(self):
         path = get_notebook_path('broken2.ipynb')
         result_path = os.path.join(self.test_dir, 'broken2.ipynb')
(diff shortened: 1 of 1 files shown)


willingc:
> Sorry to read that, didn't know that Python against innovation and effective development.

We are not against it. We are being conservative since we have seen a bunch of AI slop in the past two years. Personally, a few of us are working on adding md files for better AI context to the repo.

winniehell:
@Borda that is not an argument but only a weird analogy.

I like to use my pressure pot for potatoes but even though it has never exploded, I refuse to put in my pizza and chocolate.

also using more pressure pots doesn't directly cause fewer explosions.

Borda:
> I like to use my pressure pot for potatoes but even though it has never exploded, I refuse to put in my pizza and chocolate.
> also using more pressure pots doesn't directly cause fewer explosions.

@winniehell cool, seems you understood, and thank you for rephrasing what I just said :)

PR #860, opened 2026-01-13
Fallback to default kernel if none specified.

### LINKED ISSUE #338: Default kernel when none is provided
Right now, when calling `execute_notebook`, papermill requires either an explicitly specified `kernel_name`, or a kernelspec to be present in the notebook's metadata.

This complicates situations a) where we are writing notebooks that are compatible across the py2/py3 language barrier and b) where we are distributing notebooks to the public.

In the first case, if we specify either a `python2` or `python3` kernel in our notebooks, we will work out of the box (on a default install) on one Python version, but not on the other, even if the notebook is compatible. There are many workarounds, but all are burdensome and distracting.

In the second case, regardless of what kernel we specify in our notebooks, we are not guaranteed that it will be present in the environment to which we are distributing notebooks.

Granted that it is impossible to magically resolve these issues in general, does anyone see a strong downside to trying to infer a kernel when it is not specified, perhaps following the scheme in #262 (looking at the language), or perhaps looking at Jupyter settings such as `MappingKernelManager.default_kernel_name`?

aebrahim:
We are being overly aggressive by raising an error if a kernel cannot be detected in a notebook. In an environment with only one kernel installed, the notebook executes just fine without errors, and we should not prematurely raise an error.

Fixes #338

### DIFF
--- a/papermill/utils.py
+++ b/papermill/utils.py
@@ -40,17 +40,9 @@ def nb_kernel_name(nb, name=None):
     Returns
     -------
     str
-        The name of the kernel
-
-    Raises
-    ------
-    ValueError
-        If no kernel name is found or provided
+        The name of the kernel or an empty string if none is found
     """
-    name = name or nb.metadata.get('kernelspec', {}).get('name')
-    if not name:
-        raise ValueError("No kernel name found in notebook and no override provided.")
-    return name
+    return name or nb.metadata.get('kernelspec', {}).get('name', '')
 
 
 def nb_language(nb, language=None):


aebrahim:
Sorry to direct ping someone else - @rgbkrk

rgbkrk (approved):

rgbkrk:
Can you rebase? This branch is out of date for me to merge. I promise I'll come back 😬

aebrahim:
@rgbkrk rebased!

rgbkrk:
Alright so I'm looking at this with fresh eyes now. This is a major breaking change to the API. I would have to make a major release _just_ for this.

What is stopping your system from setting the appropriate metadata in the notebook format?

aebrahim:
We check in our notebooks to git, and they execute in many different environments, one of which is dagstermill in a "production" setup. There will always be skew for the specifics of the kernel name. For example, on developer machines, it could be arbitrary based on the venv name, and that will certainly not match the production environment.

We therefore strip the kernel metadata from the notebook when we check it in, so every environment can run it in whichever kernel they feel is appropriate. This also prevents lots of churn when multiple people with different settings edit a notebook, which causes endless merge conflicts within git. I don't think we are the only ones who have adopted this solution, because this pre-commit hook seems to be pretty widely used to me: https://github.com/srstevenson/nb-clean

Our production environment which runs papermill has only one kernel available, so falling back to the default kernel would be the behavior which makes the most sense for our use case. I think that the specific issue we face about kernel mismatched names is the same issue faced by the poster in #338

aebrahim:
Also it's not my project so my opinion on semantic versioning is fairly meaningless, but for what it's worth from my understanding, I think this would be a minor version change, not a major one. Before this PR, papermill **never** ran notebooks that did not have a version specified, rejecting them as invalid. This PR **adds** support to execute them by using the default executor, if available. Because it does not break existing supported flows and only adds a new one, I think it would be a minor version change (e.g. for `2.8.0` at this moment in time)

rgbkrk:
Thanks for that context. That helps. I really hate the mess that is arbitrary kernels per venv rather than having the notebook presume to use the current venv/project from the directory its in. I'm working on a notebook app that handles that in [nteract/desktop](https://github.com/nteract/desktop) but it doesn't solve every single place people run notebooks (VS Code, Jupyterlab, classic, etc.) I think I'll want to take your perspectives into account there too.

You've still got language info in the notebook metadata right? Just double checking as then we can rely on the default language kernel matching up with the stated language of the notebook.

rgbkrk (approved):

rgbkrk:
I'm still seeing merge conflicts with `main`. Did you rebase against `origin/main` after pulling it in?

aebrahim:
Rebased again!

aebrahim:
Thank you for spending so much time thinking about our use case.

The pre-commit hook we use also strips language out. It's not the reason we started using it, but is a consequence of the tool we use. Because we're only running with one kernel available, falling back to the default language for that kernel is the preferred behavior, but I'm sure we could try to figure out how to modify that pre-commit hook if for some reason that was not possible.

rgbkrk:
I'll have to ship this tomorrow or sometime this week after looking at everything else we've been up to as well.

PR #878, opened 2026-04-17
Debian patches for 2.7.0

plenaerts:
During debian packaging we are applying the patches in this PR to make papermill pass its test suite. The individual commits should explain themselves, but in a nutshell:
- a file was remaining after tests, we remove that artefact
- use importlib.metadata.entrypoints instead of entrypoints as stated by https://github.com/takluyver/entrypoints
- use absolute imports in test suite
- fix a number of smaller issues in tests, including pointing pyproject.toml explicitly to papermill.tests

Thanks for considering merging our suggestions.

### DIFF
--- a/papermill/engines.py
+++ b/papermill/engines.py
@@ -3,9 +3,9 @@
 import datetime
 import sys
 from functools import wraps
+from importlib.metadata import entry_points
 
 import dateutil
-from importlib.metadata import entry_points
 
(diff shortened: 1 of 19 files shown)


plenaerts:
Apparently one of the patches fixes an indent issue introduced by a previous patch. Apologies for that!

PR #892, opened 2026-06-30
Copy parameter cell metadata to injected parameters cell

### LINKED ISSUE #731: Metadata of parameters cell not copied (`{'slideshow': {'slide_type': 'skip'}`)
## 🐛 Bug

I want to export a executed notebook to a slideshow via nbconvert. I use the metadata `{'slideshow': {'slide_type': 'skip'}` on the parameter cell (hide the cell in the slideshow). The inserted cell does not copy this metadata and is visible in the slideshow export. The original parameter cell is hidden.

Looking at the code this is expected, because a fresh cell is inserted:
https://github.com/nteract/papermill/blob/3ad2dacbf539967cfed46bfa8516afbb8a120849/papermill/parameterize.py#L95)

There are two approaches I could think of fixing this:
1. Copy the metadata and omitting `'tags': ['parameters']` from the parameter cell.
2. Letting the user set the metadata of the inserted parameters cell.

I am in favor of 1. since in my opinion this is the expected behavior and does not need any additional user code.

## Workaround

Input the metadata via a post execution script into the correct location

After we decide on an approach I create a PR for this if necessary.

Thanks!
Julian

paranoa233:
## Summary
- copy metadata from the original `parameters` cell to the injected parameters cell
- filter the original `parameters` tag while preserving other metadata/tags
- keep report mode source hiding compatible with copied `jupyter` metadata

Fixes #731.

## Tests
- `python -m pytest papermill/tests/test_parameterize.py -q`
- `python -m pytest papermill/tests/test_parameterize.py papermill/tests/test_inspect.py papermill/tests/test_utils.py -q`

Note: `python -m pytest papermill/tests/test_execute.py -q` passes 25/26 locally; the remaining failure is `TestBrokenNotebook2.test`, where this minimal environment produces an `error` output instead of the expected `display_data` output for the fixture notebook. The failure is unrelated to parameter metadata handling.

### DIFF
--- a/papermill/parameterize.py
+++ b/papermill/parameterize.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 from datetime import datetime, timezone
 from uuid import uuid4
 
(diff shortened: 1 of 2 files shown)


ayushnandi718-dev (commented):
Good change. Preserving cell metadata (like slideshow type) from the original parameters cell is important for presentations. Verified the logic handles the edge case where no parameters cell exists. LGTM!

